PR #4, opened 2022-01-22
Add and handle dequeue timeout configuration

vchan:


saq/worker.py
```diff
@@ -140,32 +142,34 @@ async def monitor(self, task, job_id):
...
         try:
-            job = await self.queue.dequeue()
-            job_id = job.id
-            job.started = now()
-            job.status = Status.ACTIVE
-            job.attempts += 1
-            logger.info("Processing %s", job)
-            await job.update()
-            context = {**self.context, "job": job}
+            job = await self.queue.dequeue(self.dequeue_timeout)
```
tobymao, line 150: can we do an early return here instead?

we don't care about any call backs if there's no job right

tobymao (approved):

Issue #5, opened 2022-01-23
Mechanism for transferring state from enqueing process to worker


sondrelg:
Hi @tobymao,

The `before_process` and `after_process` hooks are great. Do you think it would also be possible to add an event hook for `on_enqueue`. 

Celery has the `before_task_publish` event - this is really what I'd like to mirror.

```python
# celery example

@before_task_publish.connect
def transfer_correlation_id(headers: Dict[str, str], **kwargs: Any) -> None:
    """
    Attach request ID to the task headers.
    """
    cid = correlation_id.get() 
    if cid:
        headers[header_key] = cid
```

I need to be able to pass state from this event handler, to the worker, so there needs to be some data structure involved that is passed to the worker. Celery uses a plain `dict` I believe.

If you think this is a good idea, I'd also be happy to open a PR for this myself 👍

tobymao:
i'm not sure I understand the use case.

when you do await queue.enqueue(...) you already have the job and so you have access to the job etc.

can you go into a bit more detail about what you need to do?

you could do

await Job(correlation_id=corellation_id.get()).enqueue()

sondrelg:
Yes I understand the confusion. 

You are right, I *could* pass the `correlation_id` to a job directly when queueing a job, but that's not what I want to do; I want to set up automated state transfer in *all* queued jobs 🎉

The correlation ID in this example is a UUID generated in a middleware, for an incoming request. I have a log filter that logs the unique ID in all logs output in the thread for the single request. 

When I queue a job, I want to automatically transfer that ID from the web-server to the worker, before:

- using the `before_process` event handler to set the unique ID to be logged for all logs output by the worker
- using the `after_process` event handler to clean up

More than that, I would really like to build this into [this package](https://github.com/snok/asgi-correlation-id) so that this whole thing just works out of the box 🙂

tobymao:
ok sounds good, perhaps you can just add a context variable to job and then do the before enqueue hook, happy to review your PR for this

tobymao:
closing this issue for now, again, happy to review a PR when it's up

PR #11, opened 2022-03-28
Add Queue.map

barakalon:
This adds the following:

1. `Job.wait_for_result` - This is a convenience that wraps `Job.refresh`, returning job result or propagating the error
2. `Queue.apply` - This is another convenience that wraps `Queue.enqueue`, enqueuing the job and immediately waiting for its results
3. `Queue.map` - This enqueues multiple jobs and then waits for all their results.
4. `Queue.cancel_scope` - This tracks all jobs started within the context manager's scope and aborts all the jobs if an exception is raised. This will be helpful for aborting a job along with all its child jobs. This is loosely inspired by [Trio's CancelScope](https://trio.readthedocs.io/en/stable/reference-core.html#trio.CancelScope).
5. `Queue.on_enqueue` - This adds a hook for wrapped job enqueueing. I believe this solves part of https://github.com/tobymao/saq/issues/5. Note, instead of a "before" and "after" hook, this allows a async generator, similar to [pytest's yield fixtures](https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended). 

Some unrelated files were accidentally reformatted when running `black .` I think that's fine.

@tobymao


saq/queue.py
```diff
@@ -393,12 +412,122 @@ async def enqueue(self, job_or_func, **kwargs):
...
+
+        jobs = [
+            await self.enqueue(job_or_func, **{**kwargs, **kw}) for kw in iter_kwargs
+        ]
+
+        results = []
+        for job in jobs:
+            wait_timeout = expires - time.time()
+            if wait_timeout <= 0:
+                raise asyncio.TimeoutError()
```
tobymao, line 494: why not use the wait_for async primitive

saq/queue.py
```diff
@@ -393,12 +412,122 @@ async def enqueue(self, job_or_func, **kwargs):
...
+        return results
+
+    @asynccontextmanager
+    async def cancel_scope(self):
+        """
+        Context manager for a "cancel scope".
 
-        logger.info("Enqueuing %s", job)
-        return job
+        This tracks all jobs enqueues within the scope and ensures that all are aborted
```
tobymao, line 505: typo enqueued

saq/queue.py
```diff
@@ -393,12 +412,122 @@ async def enqueue(self, job_or_func, **kwargs):
...
+        job = await self.enqueue(job_or_func, **kwargs)
+        return await job.wait_for_result(timeout)
+
+    async def map(self, job_or_func, iter_kwargs, timeout=None, **kwargs):
+        """
+        Enqueue multiple jobs and collect all of their results.
+
+        This raises a JobError if any of the jobs are unsuccessful.
+
+        This waits for job results in the order they are enqueued, so JobExceptions are
```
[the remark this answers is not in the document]
barakalon: I was running into issues when running large amounts of jobs (10k).

I figured it had to do with large amounts of Redis subscriptions. I didn't spend too much time debugging because I figured waiting for jobs in order was less resource intensive and probably good enough in most cases.

I can dig more into this if you feel it's necessery.

saq/queue.py
```diff
@@ -393,12 +412,122 @@ async def enqueue(self, job_or_func, **kwargs):
         job.queued = now()
         job.status = Status.QUEUED
 
-        if not await self._enqueue_script(
+        on_enqueue_generators = [
+            on_enqueue(job) for on_enqueue in self._on_enqueue.values()
+        ]
+        for gen in on_enqueue_generators:
+            await run_iterator_to_first_yield(gen)
```
tobymao, line 419: why is this done? first_yield and then later run iterator to completion?

saq/queue.py
```diff
@@ -393,12 +412,122 @@ async def enqueue(self, job_or_func, **kwargs):
...
+            wait_timeout = expires - time.time()
+            if wait_timeout <= 0:
+                raise asyncio.TimeoutError()
+
+            results.append(await job.wait_for_result(wait_timeout))
+
+        return results
+
+    @asynccontextmanager
+    async def cancel_scope(self):
```
tobymao, line 501: i wonder if the name would be more clear as a batch or something 

async with queue.batch():

saq/queue.py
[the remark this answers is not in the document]
tobymao: i see this can be solved with a semaphore

https://github.com/tobymao/saq/blob/baa37aa6938ccbc8f7e4fd81d6b9c4ffabdbf5e8/benchmarks/simple.py#L11

it doesn't make much of a difference for waiting for the results, but certainly enqueing as on line 487 is much more efficient with an async gather as opposed to for loop

if it's not too much work can we change it use a semaphore, that might also make the code cleaner because then you can just use a wait for instead of a custom time check

saq/queue.py
barakalon: This let's users write callbacks that look like this:


```python
async def on_enqueue(job):
    state = await before()
    try:
        yield
    finally:
        await after(state)
```

That makes it really easy to share state between before and after callbacks. Otherwise you have to mess with ContextVars, monkeypatching jobs, or some other messy techniques.

I don't think we strictly need any `after` callbacks yet. So maybe we should just remove this in favor of a single `before_enqueue`?

saq/queue.py
barakalon: I forget why I did it this way. Let me switch back to `wait_for`

saq/queue.py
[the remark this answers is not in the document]
tobymao: can we make subscribe more efficient by reusing the same subscription ?

saq/queue.py
barakalon: Yeah, I struggled with the name here.

`batch` is nice and simple. I'll switch to that

barakalon: Hm, more thoughts here.

`Batch` feels like an _instance_ of something to me. For example, it seems reasonable to have to simultaneous batches.

In the way it's implemented here, you can only stack these scopes.

saq/queue.py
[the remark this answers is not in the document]
barakalon: Ah yeah I like that idea.
Let me rework this.

saq/queue.py
tobymao: batch can be thought of as a verb, like batch things

saq/queue.py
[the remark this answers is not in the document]
barakalon: re: the semaphore
I think this is something that aioredis can handle for us with it's `BlockingConnectionPool`.
Then we don't need to litter our code with semaphore acquisition.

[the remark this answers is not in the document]
barakalon: Although we might need some documentation for users on how/when to use such a pool.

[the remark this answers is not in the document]
tobymao: ah interesting, can we make that the default? i didn't realize they had one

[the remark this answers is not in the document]
tobymao: sorry if it's a bit  more work, but can you test the simple bench marks with blocking connection pool so there's no degradation in perf over a semaphore?

saq/queue.py
```diff
@@ -228,12 +249,18 @@ async def listen(self, job, callback, timeout=10):
         """
         Listen to updates on job.
 
-        job: job instance
+        job: job instance or an iterable of job instances
```
tobymao, line 252: can you change it to be jobs?

barakalon: I was worried this would be a breaking change (if anyone is calling this with the kwarg `job`).

Happy to change it if you're not worried about that, though

saq/queue.py
```diff
@@ -46,6 +60,13 @@ def __init__(self, redis, name="default", dump=None, load=None):
         self._sweep = self.namespace("sweep")
         self._dump = dump or json.dumps
         self._load = load or json.loads
+        self._before_enqueue_callbacks = {}
```
tobymao, line 63: to be consistent with worker, should we call it _before_enqueue, worker has _before_process

tobymao: should we just use a set? then you can just add and remove from the set?

saq/queue.py
tobymao: we'll just bump the version, it's fine

saq/queue.py
```diff
@@ -403,3 +439,127 @@ async def enqueue(self, job_or_func, **kwargs):
...
+        timeouts: Total seconds to wait for all jobs to complete. If None or 0, wait forever.
+        kwargs: Default kwargs for all jobs. These will be overridden by those in iter_kwargs.
+        """
+        jobs = [self.prepare_job(job_or_func, **{"timeout": timeout, **kwargs, **kw}) for kw in iter_kwargs]
+        jobs_by_id = {job.id: job for job in jobs}
+
+        done = 0
+        failed_job = None
+
+        async def callback(_id, status):
```
tobymao, line 495: change to job_id

saq/queue.py
```diff
@@ -403,3 +439,127 @@ async def enqueue(self, job_or_func, **kwargs):
...
+            except JobError:
+                print("any of the jobs failed")
+
+        job_or_func: Same as Queue.enqueue
+        iter_kwargs: Enqueue a job for each item in this sequence. Each item is the same
+            as kwargs for Queue.enqueue.
+        timeouts: Total seconds to wait for all jobs to complete. If None or 0, wait forever.
+        kwargs: Default kwargs for all jobs. These will be overridden by those in iter_kwargs.
+        """
+        jobs = [self.prepare_job(job_or_func, **{"timeout": timeout, **kwargs, **kw}) for kw in iter_kwargs]
```
tobymao, line 489: was the whole factoring of prepare_jobs necessary? i think you only need the job_id  because you already do jobs_by_id and then job.refresh,

isn't it simpler just to listen to the job_id and then when you received it to get a new instance?

async def callback(job_id, status):
  job = self.job(job_id)

then you don't need to do line 490 which is the job dict, and also the prepare job on 489

tobymao: also listen takes a job, but only needs job_id, so that can be refactored

saq/queue.py
```diff
@@ -403,3 +439,127 @@ async def enqueue(self, job_or_func, **kwargs):
...
+        jobs_by_id = {job.id: job for job in jobs}
+
+        done = 0
+        failed_job = None
+
+        async def callback(_id, status):
+            nonlocal done, failed_job
```
tobymao, line 496: not super into this section

maybe it's more robust to store jobs as a set and just remove them when we receive them here as opposed to relying on a counter

for the failed job, would it be simplified if we just don't care if jobs fail? that's on the caller

saq/queue.py
barakalon: A set requires that the object is hashable. That seems like an unnecessary constraint.

As for the name, I'm cool with just `_before_enqueue`. Note, this one supports _multiple_ callbacks. That's because I'm planning to also use this functionality to support distributed tracing.

saq/queue.py
barakalon: Ah, sure, that's better.
I copy&pasted this from https://github.com/tobymao/saq/blob/master/saq/job.py#L205

saq/queue.py
barakalon: Yeah, that all makes sense.

I think I was originally trying to ensure that each job has an ID, but extracting the key generation logic should be very straightforward. I'll make this change.

saq/queue.py
barakalon: I'll switch it to a set of job ids.

Do you mean we don't care if jobs fail within this callback? Or we don't care if jobs fail within the `map` method?

saq/queue.py
tobymao: ok

saq/queue.py
tobymao: both?

tobymao: but not a fan of having to do nonlocal failed_job and also, you only track when one job is failed and not others, seems just weird, could it be cleaner to ignore failures in the map method?

barakalon: `map` is essentially the same thing as [asyncio.gather](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather).

Like `gather` returning the awaitable return values and the the tasks, `map` returns the job results and not the jobs themselves.

`gather` makes the behavior we're discussing configurable via the `return_exceptions` argument:

> If return_exceptions is False (default), the first raised exception is immediately propagated to the task that awaits on gather(). Other awaitables in the aws sequence won’t be cancelled and will continue to run.
>
> If return_exceptions is True, exceptions are treated the same as successful results, and aggregated in the result list.

Maybe we should copy asyncio and add a `return_errors` parameter?

saq/queue.py
```diff
@@ -18,14 +32,28 @@ class Queue:
...
-    def __init__(self, redis, name="default", dump=None, load=None):
+        return cls(aioredis.from_url(url), **kwargs)
+
+    def __init__(  # pylint: disable=too-many-arguments
+        self,
+        redis,
+        name="default",
+        dump=None,
+        load=None,
+        enqueue_concurrency=100,
```
tobymao, line 53: i think this is overkill, can we just have one semaphore? for this? maybe default 20 

not sure of a good name

saq/queue.py
```diff
@@ -394,12 +437,141 @@ async def enqueue(self, job_or_func, **kwargs):
...
+        iter_kwargs: Enqueue a job for each item in this sequence. Each item is the same
+            as kwargs for Queue.enqueue.
+        timeout: Total seconds to wait for all jobs to complete. If None (default) or 0, wait forever.
+        return_exceptions: If False (default), an exception is immediately raised as soon as any jobs
+            fail. Other jobs won't be cancelled and will continue to run.
+            If True, exceptions are treated the same as successful results and aggregated in the result list.
+        kwargs: Default kwargs for all jobs. These will be overridden by those in iter_kwargs.
+        """
+        iter_kwargs = [
+            {"timeout": timeout, "key": kwargs.get("key") or uuid1(), **kwargs, **kw}
```
tobymao, line 501: sorry to be picky here, but i don't like that we have to understand to do or uuid1,

wouldn't it be easier to just do Job(key).id

saq/queue.py
```diff
@@ -394,12 +437,141 @@ async def enqueue(self, job_or_func, **kwargs):
...
+            for kw in iter_kwargs
+        ]
+        job_ids = [Job.id_from_key(key["key"]) for key in iter_kwargs]
+        pending_job_ids = set(job_ids)
+
+        async def callback(job_id, status):
+            nonlocal pending_job_ids
```
tobymao, line 508: is this necessary? you aren't assigng it, so reference should be ok

saq/queue.py
barakalon: `Job` has a required `function` attribute. We can set it to some dummy value, but that also felt ugly.

saq/queue.py
barakalon: Ah, you're correct. Just leftover crud from when this was a counter. Good catch.

saq/queue.py
barakalon: Happy to switch to one, as long as we don't have any issues with reentrance. I don't think we have any such cases yet, but we'll have to keep an eye out.

As for an alternative name.... how about `max_concurrent_ops`?

benchmarks/simple.py
```diff
@@ -8,6 +8,7 @@
 SEM = asyncio.Semaphore(20)
```
tobymao, line 8: can we remove this semaphore?

saq/queue.py
tobymao: can we have the job iniatilizer use the same helper function and add uuid1() into it as default?

saq/queue.py
tobymao: perfect

saq/queue.py
tobymao: yea i'm not sure exactly how to do this

benchmarks/simple.py
barakalon: seems like this is used by the arq benchmark?

saq/queue.py
barakalon: Seems sketchy to have implicit default logic in the Job constructor, since that's part of the public API and it might make constructing Jobs more confusing.

We could just add a `get_default_key` function in the job module 🤷 

We could also add a `prepare_job` method 😛

benchmarks/simple.py
tobymao: sounds good

saq/queue.py
tobymao: ok, yea let's add a job default key

tobymao (approved):

PR #13, opened 2022-04-01
Handle alternative deserialization in UI

barakalon:
This fixes a couple bugs:

**Alternative job serialization in the UI**
The UI would always attempt to use JSON serialization for jobs. 
This would break the UI if configuring a different dump/load function (e.g. pickle).

Now, the UI uses the same dump/load function that the worker/queue are configured with. Note - this means the UI will break if attempting to view jobs from _other_ queues that might have a different dump/load function. 

I think it's possible to solve the above problem by having two different serialization steps. Something like this:
```
Job --marshal--> JobMessage --serialize--> bytes
```
Where:
- `Job` is a user-friendly object exposed in the public API
- `marshal` are the user provided dump/load functions, but now they only apply to kwargs and result
- `JobMessage` is a simpler object that can be handled without unmarshaling. We could implement UI monitoring/control based on this object.
- `serialize` would be json.loads. And maybe something like msgpack in the future.
- `bytes` would be what we actually store in Redis

But the coupling of Queue and Job is pretty strong right now, and it gets pretty messy to untangle. So I think we should punt on all that for now.

**Viewing a job's queue in the UI**
Queue's were always binding themselves to jobs they read.
That ultimately meant jobs in the UI would always show as being part of whatever queue the web server was running off of.

I've added this `ImpersonatorQueue` to handle this.

Kind ugly.. but a pure solution involves the same proposal above.

@tobymao

### DIFF
--- a/saq/monitor.py
+++ b/saq/monitor.py
@@ -1,0 +1,123 @@
+import json
+
+from saq import Queue, Job
+from saq.queue import STATS_KEY
+from saq.utils import now
+
+
+class ImpersonatorQueue(Queue):
+    """
+    Queue that can read jobs from other queues.
+
+    This little hack is intended solely for monitoring purposes.
+    """
+
+    def impersonate(self, name):
+        return impersonate(queue=self, new_name=name)
+
+    def _init_job(self, job_dict):
+        queue_name = job_dict.pop("queue")
+        queue = self.impersonate(name=queue_name)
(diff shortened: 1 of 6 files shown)


tobymao:
discussed that this is not the correct solution, instead the web by default should only read and serve the queue it started with. 

as a convenience, we should add a way to pass multiple queue settings to the web worker

PR #14, opened 2022-04-01
Support synchronous job functions

barakalon:
This lets users define synchronous job functions.

This is essentially the same thing that FastAPI does: https://github.com/tiangolo/fastapi/blob/master/fastapi/routing.py#L152

This isn't strictly required, as it's easy to workaround for users. But IMO, this is a nice framework convenience.

@tobymao

### DIFF
--- a/saq/worker.py
+++ b/saq/worker.py
@@ -3,6 +3,7 @@ import logging
 import signal
 import traceback
 import os
+from functools import partial, wraps
 
 from croniter import croniter
 
@@ -80,10 +81,6 @@ class Worker:
                 name, function = function
             else:
                 name = function.__qualname__
-
-            assert asyncio.iscoroutinefunction(
-                function
-            ), f"{function} is not a coroutine"
 
             self.functions[name] = function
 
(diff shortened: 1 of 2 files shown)


saq/worker.py
```diff
@@ -240,6 +236,20 @@ def _process(self, previous_task=None):
             new_task.add_done_callback(self._process)
 
 
+def ensure_coroutine_function(func):
+    if asyncio.iscoroutinefunction(func):
+        return func
+
+    @wraps(func)
```
tobymao, line 243: possible to simplify this? do we need wraps because we don't use a decorator, also partial?

async def wrapped(*args, **kwargs):
     return await loop.run_in_executor(executor=None, func=lambda: func(*args, **kwargs)
 return wrapped

barakalon: Ah sure, I can switch to lambda. 

This is a decorator, it's just not invoked with the `@` syntax. But I guess that means its unlikely someone will have a direct handle on the wrapped function, so I'll remove it.

PR #15, opened 2022-04-01
Monitor multiple queues in the Web UI

barakalon:
This is the result of our discussion on https://github.com/tobymao/saq/pull/13

Before this PR, a web UI instance monitors all queues in the same Redis database. This had some fundamental issues.

This change makes it so a web UI instance only monitors the queue for the worker it's attached to.

If users want to monitor multiple queues in the web UI, they can append `--additional-web-settings` parameters to the `saq` command.

@tobymao

### DIFF
--- a/saq/__main__.py
+++ b/saq/__main__.py
@@ -26,7 +26,14 @@ def main():
     parser.add_argument(
         "--web",
         action="store_true",
-        help="Start web app",
+        help="Start web app. "
+        "By default, this only monitors the current worker's queue. To monitor multiple queues, see '--web-settings'",
+    )
+    parser.add_argument(
+        "--additional-web-settings",
+        "-a",
+        action="append",
+        help="Additional worker settings to monitor in the web app",
     )
     parser.add_argument(
         "--port",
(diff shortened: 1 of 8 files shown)


barakalon:
This has some breaking changes:

1. There is no more `saq:stats` key - it's now namespaced to each queue.
2. `Queue.info` now only returns info for _that_ queue
3. There are no more `[/api]/jobs/*` endpoints. We now need to know the queue, so they all become `[/api]/queues/{queue}/jobs/*`

tobymao:
@vchan when we add this, we need to update kube

saq/__main__.py
```diff
@@ -26,7 +26,14 @@ def main():
     parser.add_argument(
         "--web",
         action="store_true",
-        help="Start web app",
+        help="Start web app. "
+        "By default, this only monitors the current worker's queue. To monitor multiple queues, see '--web-settings'",
+    )
+    parser.add_argument(
+        "--additional-web-settings",
```
tobymao, line 33: web-settings vs additional-web-settings as stated in 30

tobymao: what about extra_settings, or monitor_settings, or just web_settings is fine

barakalon: I think every word in "additional-web-settings" adds information:
settings: settings path
web: only for web
additional: in _addition_ to the main settings

But I'll settle for `web_settings` :)

tobymao (approved):
merge whenever you're ready, great work

saq/__main__.py
tobymao: up to you, you can make the decision

PR #16, opened 2022-04-02
Add retry delay

barakalon:
This adds optional retry delay to jobs.

This also supports exponential backoff.

@tobymao

### DIFF
--- a/saq/job.py
+++ b/saq/job.py
@@ -177,7 +177,6 @@ class Job:
             or (self.heartbeat and seconds(current - self.touched) > self.heartbeat)
         )
 
-    @property
     def next_retry_delay(self):
         if self.retry_backoff:
             return exponential_backoff(
--- a/saq/job.py
+++ b/saq/job.py
@@ -63,11 +63,11 @@ class Job:
         retries: the maximum number of attempts to retry a job, defaults to 1
         ttl: the maximum time in seconds to store information about a job including results, defaults to 600
         retry_delay: seconds to delay before retrying the job
-        retry_backoff: if true, use exponential backoff for retry delays.
+        retry_backoff: If true, use exponential backoff for retry delays.
             The first retry will have whatever retry_delay is.
             The second retry will have retry_delay*2. The third retry will have retry_delay*4. And so on.
-        retry_backoff_max: if set, this caps the retry delays calculated by retry_backoff
-        retry_jitter: if true, this randomizes the retry delay calculate by retry backoff between 0 and the calculated retry delay
+            This always includes jitter, where the final retry delay is a random number between 0 and the calculated retry delay.
+            If retry_backoff is set to a number, that number is the maximum retry delay, in seconds.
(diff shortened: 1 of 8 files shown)


saq/worker.py
```diff
@@ -214,7 +214,7 @@ async def process(self):
             logger.error(error)
 
             if job:
-                if job.attempts >= job.retries:
```
barakalon, line 217: I think this was always wrong. I think this was implemented as if "retries" was actually "max_attempts".

Changing this corrects the semantics. But it's also a breaking change in behavior. Let me know if that's OK.

saq/job.py
```diff
@@ -84,6 +89,10 @@ class Job:
     heartbeat: int = 0
     retries: int = 1
     ttl: int = 600
+    retry_delay: float = 0.0
+    retry_backoff: bool = False
```
tobymao, line 93: i think there are too many args

what do you think about combining retry_backoff and max

retry_backoff: float = 0.0 means no backoff
retry_backoff=10.0 means 10 second max, no need to support infinite retry

retry_jitter... is this really necessary?

saq/utils.py
```diff
@@ -16,3 +18,26 @@ def millis(s):
 
 def seconds(ms):
     return ms / 1000
+
+
+def exponential_backoff(
+    attempts: int,
```
tobymao, line 24: sorry for this nit but the types here aren't consistent with the rest of the codebase, can you remove?

saq/worker.py
tobymao: breaking changes are always ok.

you're right, in my mind, retries means max_attempts.

and so job.attempts >= job.retries is correct because retries = 1 means that it will retry once
retries 2 means it will retry twice.

do you think this really needs to be changed?

tobymao: i think that we just have different interpretations of the word "retries" and so the current implementation is what i intended

saq/utils.py
barakalon: Ah, whoops. I had copy and pasted this from something else I wrote and left these in.

saq/job.py
barakalon: > retry_jitter... is this really necessary?

Backoff is kind of pointless without jitter: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

How about:

```
retry_backoff: Union[bool, float] = False
```

False means no backoff.
True means backoff with no max.
A number means backoff with a max.
And we always have jitter.

saq/worker.py
barakalon: _re_try means trying _again_ haha.

But that's OK. If that's how you interpret it, I can change it back.

saq/job.py
tobymao: perfect, i'd rather always have jitter

and we can do false, true, and float

tobymao (approved):

PR #18, opened 2022-04-08
Better support for context propagation

barakalon:
This should be able to handle most cases of context propagation between jobs.

This doesn't have a strong opinion on _what_ context is propagated and instead just allows for arbitrary metadata. This should work for OpenTracing, DataDog tracing, asgi-correlation-id, etc.

@tobymao 
@sondrelg

### DIFF
--- a/saq/job.py
+++ b/saq/job.py
@@ -70,6 +70,7 @@ class Job:
             If retry_backoff is set to a number, that number is the maximum retry delay, in seconds.
         scheduled: epoch seconds for when the job should be scheduled, defaults to 0 (schedule right away)
         progress: job progress 0.0..1.0
+        meta: arbitrary metadata to attach to the job
     Framework Set Properties
         attempts: number of attempts a job has had
         completed: job completion time epoch seconds
@@ -101,6 +102,7 @@ class Job:
     result: typing.Any = None
     error: typing.Optional[str] = None
     status: Status = Status.NEW
+    meta: dict = dataclasses.field(default_factory=dict)
 
     def __repr__(self):
         kwargs = ", ".join(
@@ -119,6 +121,7 @@ class Job:
                 "result": self.result,
                 "error": self.error,
                 "status": self.status,
+                "meta": self.meta,
             }.items()
             if v is not None
         )
@@ -140,11 +143,21 @@ class Job:
         return f"{ABORT_ID_PREFIX}{self.key}"
 
     def to_dict(self):
-        return {
-            k: v.name if k == "queue" else v
-            for k, v in self.__dict__.items()
-            if v != self.__dataclass_fields__[k].default  # pylint: disable=no-member
-        }
+        result = {}
+        for field in dataclasses.fields(self):
+            value = getattr(self, field.name)
+
+            if field.name == "queue" and value:
+                value = value.name
+
+            default = field.default
+            if field.default_factory != dataclasses.MISSING:
+                default = field.default_factory()
+
+            if value != default:
+                result[field.name] = value
+
+        return result
 
     def duration(self, kind):
         """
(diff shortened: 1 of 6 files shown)


saq/job.py
```diff
@@ -140,11 +143,21 @@ def abort_id(self):
         return f"{ABORT_ID_PREFIX}{self.key}"
 
     def to_dict(self):
-        return {
-            k: v.name if k == "queue" else v
-            for k, v in self.__dict__.items()
-            if v != self.__dataclass_fields__[k].default  # pylint: disable=no-member
-        }
+        result = {}
```
barakalon, line 146: Take a critical look here.

The `meta` attribute has a default factory so that it's easy for different plugins to tack on various metadata without having to worry about `None`.

But that makes the logic for removing defaults a little bit more complex. With the way this is currently implemented, we'll end up generating a frivolous UUID for the key every time we call `to_dict`.

I think what we actually want is something like pydantic's `exclude_unset` option, but I don't think that's so easy with dataclasses. At least with how we're using them.

saq/worker.py
```diff
@@ -83,6 +84,14 @@ def __init__(
 
             self.functions[name] = function
 
+    async def _before_process(self, ctx):
```
barakalon, line 87: Secretly, these are intended to be nice monkey-patching points for dd-trace-py.

saq/worker.py
```diff
@@ -219,11 +227,10 @@ async def process(self):
...
                     await job.retry(error)
         finally:
-            if job:
-                self.job_task_contexts.pop(job)
+            if context:
+                self.job_task_contexts.pop(job, None)
 
-                if self.after_process and job.completed:
-                    await self.after_process(context)
+                await self._after_process(context)
```
barakalon, line 233: This is a breaking change - 

This always calls `after_process`, even if not completed

saq/job.py
tobymao: i wonder if we should just simplify and remove the pruning of default values

saq/worker.py
tobymao: i figured as much, but why not just pass it in instead of overriding?

saq/job.py
barakalon: Yeah, adding defaults in the serialized job feels cleaner.

I think the main downside would be Redis memory overhead and bandwidth, right?

saq/worker.py
barakalon: ddtrace loves it's monkeypatching: https://ddtrace.readthedocs.io/en/stable/integrations.html

But that's of course pretty fragile.

I figured adding these methods was a reasonable refactor anyways, and we could make the ddtrace decision later.

Happy to remove these if we think it's ugly, though.

barakalon: Also, if we opt for passing in, we should make it possible to have multiple `before_process` and `after_process` callbacks

saq/worker.py
tobymao: @vchan will this cause a problem with our current nerf settings that auto commits? @barakalon suggests we rollback instead

barakalon: Yeah, I think it's a best practice to explicitly commit on success. 

Also, I don't know how long nerf jobs are supposed to run, but in general, I'd recommend keeping database transactions _short_. If you begin a transaction, then do a bunch of work not related to that transaction, _then_ finally commit when the job completes, you might cause problems for the database. Especially if that transaction is locking any objects.

tobymao: oooooo didn't realize about the transaction, that's a good thought

tobymao (approved):

sondrelg:
I've finally gotten around to implementing this myself. It looks great @barakalon! Will add saq-docs to asgi-correlation-id shortly in https://github.com/snok/asgi-correlation-id/pull/47 👍

PR #20, opened 2022-04-09
continue to prune dict

tobymao:
@barakalon wasn't that hard to continue pruning results, default for the dataclass just returns MISSING_VALUE which is fine.


barakalon:
Ah, sure. This is similar to what I had before, it just handles “meta” specifically instead of generically via the default_factory.
👍

PR #21, opened 2022-04-13
Use queue name in the job key

acreux:
cron jobs may not be unique if they have the same name but run in different queues

### DIFF
--- a/saq/queue.py
+++ b/saq/queue.py
@@ -505,7 +505,7 @@ class Queue:
             }
             for kw in iter_kwargs
         ]
-        job_ids = [Job.id_from_key(key["key"]) for key in iter_kwargs]
+        job_ids = [Job.id_from_key(key["key"], queue_name=self.name) for key in iter_kwargs]
         pending_job_ids = set(job_ids)
 
         async def callback(job_id, status):
--- a/saq/queue.py
+++ b/saq/queue.py
@@ -505,7 +505,9 @@ class Queue:
             }
             for kw in iter_kwargs
         ]
-        job_ids = [Job.id_from_key(key["key"], queue_name=self.name) for key in iter_kwargs]
+        job_ids = [
+            Job.id_from_key(key["key"], queue_name=self.name) for key in iter_kwargs
+        ]
         pending_job_ids = set(job_ids)
 
         async def callback(job_id, status):
--- a/saq/queue.py
+++ b/saq/queue.py
@@ -260,16 +260,16 @@ class Queue:
                     logger.info("Sweeping job %s", job)
         return swept
 
-    async def listen(self, job_ids, callback, timeout=10):
+    async def listen(self, job_keys, callback, timeout=10):
         """
         Listen to updates on jobs.
 
-        job_ids: sequence of job ids
-        callback: callback function, if it returns truthy, break
+        job_keys: sequence of job keys
+        callback: callback function, if it returns truthy, break. Takes (job_id, status) as arguments
         timeout: if timeout is truthy, wait for timeout seconds
         """
         pubsub = self.redis.pubsub()
-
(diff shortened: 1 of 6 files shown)


barakalon:
What's the intent here? UUIDs should be globally unique, so I'm assuming there is some issue with an application setting non-unique keys?

acreux:
we had a conflict with 2 same cron jobs in 2 different queues

barakalon:
Maybe we should include the queue name here? https://github.com/tobymao/saq/blob/master/saq/worker.py#L134

tobymao:
it's easier if uid's are locally unique, do you have a use case where it needs to be global?

barakalon:
Talked offline. Conclusion:
We should namespace job ids by queue, as this PR does.
But we should change `Queue.job` (and any similar methods) to lookup jobs by `key`, not `id`.

saq/queue.py
```diff
@@ -384,7 +385,8 @@ async def dequeue(self, timeout=0):
                 "BLMOVE", self._queued, self._active, "RIGHT", "LEFT", timeout
             )
         if job_id is not None:
-            return await self.job(job_id)
+            job_key = Job.key_from_id(job_id.decode())
```
acreux, line 388: not a fan of getting job_key from `job_id` here with the `decode` and the `split` in `Job.key_from_id`

saq/queue.py
```diff
@@ -260,16 +260,16 @@ async def sweep(self, lock=60):
...
 
-        job_ids: sequence of job ids
-        callback: callback function, if it returns truthy, break
+        job_keys: sequence of job keys
+        callback: callback function, if it returns truthy, break. Takes (job_id, status) as arguments
         timeout: if timeout is truthy, wait for timeout seconds
         """
         pubsub = self.redis.pubsub()
-
+        job_ids = [Job.id_from_key(job_key, self.name) for job_key in job_keys]
```
tobymao, line 272: i don't like having to do id_from_key, can we remove this method? can't we just do something like

queue.job_id(job_key)

because you can never call Job.id_from_key without the queue anyways, so forget the Job and call the queue

saq/queue.py
```diff
@@ -505,8 +507,12 @@ async def map(
...
-        job_ids = [Job.id_from_key(key["key"]) for key in iter_kwargs]
-        pending_job_ids = set(job_ids)
+        job_keys = [key["key"] for key in iter_kwargs]
+        pending_job_keys = set(job_keys)
+        pending_job_ids = set(
+            Job.id_from_key(job_key, queue_name=self.name)
+            for job_key in pending_job_keys
+        )
 
         async def callback(job_id, status):
```
barakalon, line 517: If `listen` uses job keys as a parameter, it should pass job keys to the callback. In other words, I think we should extract the key out of the id before calling this callback. That would make this `map` method simpler.

saq/queue.py
tobymao: let's not do this, we should just have a second function or call job with directly 

job(id=id) which bypasses the logic

def job(key=None, id=None) something like this?

saq/queue.py
tobymao: @barakalon thoughts?

saq/queue.py
barakalon: How about a private `_get_job_by_id` (or something like that) method that expects `id`.
The public `job` method can just wrap that method, and this method can call it directly.

barakalon (commented):
While we're making breaking changes, what if we swapped `key` and `id`?

I.e., `id` is the job identifier that users can override. `key` is the Redis key.

That might be a refactor for another day, though.

saq/queue.py
barakalon: Makes sense to me 👍

tobymao (approved):

barakalon (approved):

PR #27, opened 2022-05-26
Fix import setting module

aprilahijriyan:


Nimond (approved):

Nimond:
Thanks a lot!

saq/worker.py
```diff
@@ -280,6 +281,7 @@ def import_settings(settings):
 
 
 def start(settings, web=False, extra_web_settings=None, port=8080):
+    sys.path.insert(0, os.getcwd())  # fix: https://github.com/tobymao/saq/issues/26
```
barakalon, line 284: This feels risky to me. We don't actually know what directory saq is being invoked from.

_If_ we add something like this, it should probably be in https://github.com/tobymao/saq/blob/master/saq/__main__.py

aprilahijriyan:
I agree with @barakalon suggestion. How about you @tobymao ?

barakalon:
To be clear, I'm leaning heavily _away_ from this.

tobymao:
@aprilahijriyan as i said it’s best if you fix your app or add a python path, this is how arq works as well fyi

PR #29, opened 2022-05-27
Allow users to expire, keep, or disable storage of job information

paul-finary:
Currently, the information attached to a job that has been processed are necessarily stored back in redis. In some cases, this is not needed, and unwanted.

I added a way for users to disable this feature by making the ttl have three states:
- positive (e.g. 600): the job is saved back to redis with an expiry after processing
- neutral (e.g. 0): the job is saved back to redis without expiry
- negative (e.g. -1): the job is not saved back to redis

I thought about using positive, None, and 0, but decided to stay with one type, not sure what's best honestly.

### DIFF
--- a/saq/queue.py
+++ b/saq/queue.py
@@ -374,6 +374,8 @@ class Queue:
                 pipe = pipe.setex(job_id, job.ttl, self.serialize(job))
             elif job.ttl == 0:
                 pipe = pipe.set(job_id, self.serialize(job))
+            else:
+                pipe.delete(job_id)
 
             await pipe.execute()
 
--- a/saq/queue.py
+++ b/saq/queue.py
@@ -370,9 +370,9 @@ class Queue:
         async with self.redis.pipeline(transaction=True) as pipe:
             pipe = pipe.lrem(self._active, 1, job_id).zrem(self._incomplete, job_id)
 
-            if job.ttl:
+            if job.ttl > 0:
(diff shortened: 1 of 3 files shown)


saq/queue.py
```diff
@@ -370,9 +370,9 @@ async def finish(self, job, status, *, result=None, error=None):
         async with self.redis.pipeline(transaction=True) as pipe:
             pipe = pipe.lrem(self._active, 1, job_id).zrem(self._incomplete, job_id)
 
-            if job.ttl:
+            if job.ttl > 0:
                 pipe = pipe.setex(job_id, job.ttl, self.serialize(job))
-            else:
+            elif job.ttl == 0:
```
barakalon, line 375: Should we be deleting the key if job.ttl < 0? 

We store the serialized job when enqueueing it, so doing nothing might actually leave the job stored forever.

barakalon (commented):
I dig -1.

Can you add a test for this?

saq/queue.py
tobymao: yea, we need to explicitly delete the job

paul-finary: Alright, I fixed it

paul-finary:
> I dig -1.
> 
> Can you add a test for this?

Just added several tests for the 3 different ttl "states"

tobymao (approved):

PR #32, opened 2022-06-05
tests(worker): added additional test cases around long-running jobs

cofin:
After using this library for a few weeks, there have been a few times where the default timeout of 10 early terminated a few of my jobs.  I've added a few additional test cases documenting the expected outcomes of "long running jobs" with various timeout configurations.

### DIFF
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -24,6 +24,11 @@ async def sleeper(ctx):
     return {"a": 1, "b": []}
 
 
+async def long_sleeper(ctx):
+    await asyncio.sleep(ctx.get("sleep", 15))
+    return {"a": 1, "b": []}
+
+
 async def cron(_ctx):
     return 1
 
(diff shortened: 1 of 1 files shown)


barakalon (commented):
I love the attention to tests!

I'm a bit worried that this adds 10s+15s+15s=40s to the overall unit test duration. We try to keep the unit tests fast, as tests can be part of a developer's iteration cycle.

What behavior do these tests add _beyond_ the "test_sleeper" tests?

cofin:
Thanks for the quick feedback.  The initial purpose of these tests were for me to better understand how the various timeout configurations impacted my active jobs.

In my particular use case, I have mostly long(ish) running ETL jobs.  Initially, I was just using `enqueue` without supplying a change to the default timeout setting, and I found that my jobs were terminated by the default Job.timeout.

With all of that being said, perhaps the test cases are a bit redundant compared to the existing timeout jobs.  Alternately, this could be something I could contribute to the documentation/examples to help clarify standard behavior.

As part of the project I'm using this in, I've also added a `monitored_task` decorator that adds a periodic heartbeat while a task is executing.  I'll be pushing that over in a bit.  If there are other areas where help is needed, let me know.

tobymao:
thanks for the pr, but there no need for these tests, happy to accept another PR with more robust documentation

PR #33, opened 2022-06-05
Periodic task decorator

cofin:
I've been using this decorator to enable jobs to automatically execute `update` on an interval.

The decorator determines if the job has a heartbeat setting enabled, and if so, it automatically ticks while the job is executing.

The current formula for how often to tick is the following:  `max(round(job.heartbeat/2),1)`.  

I've added an example of how it works in the `examples/monitored.py`, and I took a shot at test cases.

```python
from saq.job import monitored_job

@monitored_job
async def monitored_sleeper(ctx):
    await asyncio.sleep(ctx.get("sleep", 10))
    return {"a": 1, "b": []}
```

The heartbeat will automatically start once you've scheduled a job for this function with `job.heartbeat` 

```python
await queue.apply(
        func,
        heartbeat=3,
    )
```

After starting a job with the decorator applied, you'll see that it's executing periodically in the logs:
```
INFO:saq:Processing Job<function=monitored_cron_job, queue=default, id=saq:job:default:cron:monitored_cron_job, scheduled=1654467040, progress=0.0, start_ms=16107, attempts=1, status=active, meta={}>
INFO:saq:starting heartbeat service for saq:job:default:cron:monitored_cron_job. Ticking every 1 second(s)
excuting cron job
INFO:saq:ticking heartbeat for saq:job:default:cron:monitored_cron_job, and sleeping for 1 second(s)
INFO:saq:stopping heartbeat service for saq:job:default:cron:monitored_cron_job
```

### DIFF
--- a/examples/monitored.py
+++ b/examples/monitored.py
@@ -10,7 +10,7 @@ logger.setLevel(logging.DEBUG)
 
 @monitored_job
 async def monitored_sleeper(ctx):
-    print("executing monitored sleeper job.")
+    logger.info("executing monitored sleeper job.")
 
     await asyncio.sleep(30)
     return {"a": 6}
@@ -18,19 +18,19 @@ async def monitored_sleeper(ctx):
 
 async def sleeper(ctx):
     await asyncio.sleep(30)
-    print("executing regular sleeper job.  I should fail due to missed heartbeat")
+    logger.info("executing regular sleeper job.  I should fail due to missed heartbeat")
 
     return {"a": 6}
 
 
 @monitored_job
 async def monitored_cron_job(ctx):
-    print("excuting cron job")
+    logger.info("excuting cron job")
     await asyncio.sleep(6)
 
 
 async def cron_job(ctx):
-    print("excuting regular cron job.  I should fail due to missed heartbeat")
+    logger.info("excuting regular cron job.  I should fail due to missed heartbeat")
     await asyncio.sleep(6)
 
 
@@ -51,12 +51,11 @@ async def enqueue(func, **kwargs):
         func,
         heartbeat=3,
     )
-    print(result)
+    logger.info(result)
 
 
 if __name__ == "__main__":
     now = time.time()
     asyncio.run(enqueue("monitored_sleeper"))
     asyncio.run(enqueue("sleeper"))
-
-    print(time.time() - now)
+    logger.info(time.time() - now)
--- a/examples/monitored.py
+++ b/examples/monitored.py
@@ -1,0 +1,62 @@
+import asyncio
+import logging
+import time
+
+from saq import CronJob, Queue, monitored_job
+
+logger = logging.getLogger("saq")
+logger.setLevel(logging.DEBUG)
+
+
+@monitored_job
+async def monitored_sleeper(ctx):
+    print("executing monitored sleeper job.")
+
+    await asyncio.sleep(30)
+    return {"a": 6}
+
+
+async def sleeper(ctx):
+    await asyncio.sleep(30)
+    print("executing regular sleeper job.  I should fail due to missed heartbeat")
+
+    return {"a": 6}
+
+
+@monitored_job
+async def monitored_cron_job(ctx):
+    print("excuting cron job")
+    await asyncio.sleep(6)
+
+
+async def cron_job(ctx):
+    print("excuting regular cron job.  I should fail due to missed heartbeat")
+    await asyncio.sleep(6)
+
+
+settings = {
+    "functions": [monitored_sleeper, monitored_cron_job, sleeper, cron_job],
+    "concurrency": 100,
+    "timers": {"sweep": 5},
+    "cron_jobs": [
+        CronJob(monitored_cron_job, cron="* * * * * */20", heartbeat=2, timeout=None),
+        CronJob(cron_job, cron="* * * * * */20", heartbeat=2, timeout=None),
+    ],
+}
+
+
+async def enqueue(func, **kwargs):
+    queue = Queue.from_url("redis://localhost")
+    result = await queue.apply(
+        func,
+        heartbeat=3,
+    )
+    print(result)
+
+
+if __name__ == "__main__":
+    now = time.time()
+    asyncio.run(enqueue("monitored_sleeper"))
+    asyncio.run(enqueue("sleeper"))
+
+    print(time.time() - now)
--- a/saq/job.py
+++ b/saq/job.py
@@ -270,7 +270,7 @@ class Job:
             setattr(self, field, getattr(job, field))
 
 
-class PeriodicHeartbeat(object):
+class PeriodicHeartbeat:
     """Class to manage agent Heartbeats while Job is running"""
 
     def __init__(self, job: Job) -> None:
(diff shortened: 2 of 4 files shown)


tests/test_worker.py
```diff
@@ -122,6 +134,47 @@ async def test_sleeper(self):
...
+        swept = await self.queue.sweep()
+        self.assertNotIn(
+            set(swept),
+            {job.id.encode()},
+        )
+        await task
+        await job.refresh()
+        self.assertEqual(job.status, Status.COMPLETE)
+
+    async def test_missed_heartbeat(self):
```
barakalon, line 158: There's also this test: https://github.com/tobymao/saq/blob/master/tests/test_queue.py#L235

That test is a bit more "white box" compared to this one.

But do we need both?

examples/monitored.py
```diff
@@ -0,0 +1,62 @@
...
+
+from saq import CronJob, Queue, monitored_job
+
+logger = logging.getLogger("saq")
+logger.setLevel(logging.DEBUG)
+
+
+@monitored_job
+async def monitored_sleeper(ctx):
+    print("executing monitored sleeper job.")
```
barakalon, line 13: We create a "logger" above, but then use print.

Maybe we should be using the logger?

examples/monitored.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    print(result)
+
+
+if __name__ == "__main__":
+    now = time.time()
+    asyncio.run(enqueue("monitored_sleeper"))
+    asyncio.run(enqueue("sleeper"))
+
+    print(time.time() - now)
```
barakalon, line 62: I think this is copy/pasted from another example, which is more concerned about latency. But I don't think this elapsed time means much for this example.

saq/job.py
```diff
@@ -179,8 +187,10 @@ def _duration(self, a, b):
     def stuck(self):
         """Checks if an active job is passed it's timeout or heartbeat."""
         current = now()
+        # if timeout == None, the if statement will fail for
```
barakalon, line 190: Is this a comment for _why_ you added the condition below? Good catch, but I don't think this comment should be in the code itself.

saq/job.py
```diff
@@ -258,3 +268,77 @@ def replace(self, job):
         """Replace current attributes with job attributes."""
         for field in job.__dataclass_fields__:
             setattr(self, field, getattr(job, field))
+
+
+class PeriodicHeartbeat:
+    """Class to manage agent Heartbeats while Job is running"""
+
+    def __init__(self, job: Job) -> None:
```
barakalon, line 276: saq isn't using type annotations yet. I don't think they add much without using something like `mypy` in regular checks.

Can you remove this for now?
For documentation/IDE hints, use sphinx docstrings (as you do below).

saq/job.py
```diff
@@ -258,3 +268,77 @@ def replace(self, job):
...
+        self._running: bool = True
+        self._heartbeat_task: typing.Optional[asyncio.Task] = None
+        self.job = job
+        self.heartbeat_enabled: bool = job.heartbeat and job.heartbeat > 0
+        self.time_between_ticks = (
```
barakalon, line 281: I think "tick" has a specific meaning in other contexts. Should "tick" be "beat"?

saq/job.py
```diff
@@ -258,3 +268,77 @@ def replace(self, job):
...
+
+    async def start(self, func: typing.Awaitable):
+        """start
+
+        Executes function with a periodic heartbeat.
+        Args:
+            func (typing.Awaitable): _description_
+
+        Returns:
+            _type_: _description_
```
barakalon, line 293: What is this syntax?

saq/job.py
```diff
@@ -258,3 +268,77 @@ def replace(self, job):
...
+        self.time_between_ticks = (
+            max(round(job.heartbeat / 2), 1) if self.heartbeat_enabled > 0 else 0
+        )
+
+    async def start(self, func: typing.Awaitable):
+        """start
+
+        Executes function with a periodic heartbeat.
+        Args:
+            func (typing.Awaitable): _description_
```
barakalon, line 290: I don't think this is a "func". This is the coroutine object itself.

saq/job.py
```diff
@@ -258,3 +268,77 @@ def replace(self, job):
...
+            _type_: _description_
+        """
+        if self.heartbeat_enabled:
+            logger.info(
+                "starting heartbeat service for %s. Ticking every %s second(s)",
+                self.job.id,
+                self.time_between_ticks,
+            )
+        self._heartbeat_task = asyncio.create_task(self._periodically_publish())
+        executed_func: typing.Optional[typing.Any] = None
```
barakalon, line 302: Similar comment as above.  This is the result of the coroutine, not the "executed function".

saq/job.py
```diff
@@ -258,3 +268,77 @@ def replace(self, job):
...
+        try:
+            executed_func = await func
+            self._heartbeat_task.cancel()
+            await self._heartbeat_task
+        except asyncio.CancelledError:
+            pass
```
barakalon, line 308: This will catch _all_ CancelledErrors, even those that might've been raised elsewhere (for example, if the worker is stopped: https://github.com/tobymao/saq/blob/master/saq/worker.py#L127).

I think that could potentially silence exceptions and make the worker think the job completed successfully. No bueno.

saq/job.py
```diff
@@ -258,3 +268,77 @@ def replace(self, job):
...
+                "ticking heartbeat for %s, and sleeping for %s second(s)",
+                self.job.id,
+                self.time_between_ticks,
+            )
+            await self.job.update()
+            await asyncio.sleep(self.time_between_ticks)
+
+
+# decorator to perform a heartbeat in the background while a function is running
+def monitored_job(func):
```
barakalon, line 331: "monitored_job" feels like a misnomer.

This doesn't mean that the job is now monitored... all jobs are "monitored". 

Is there a better name here?

barakalon (commented):
This strategy trusts that the job itself won't get stuck. I think this would mostly be useful for detecting crashed workers that have _also_ failed to abort jobs. 

Heartbeats can also be used to detect that the job procedure itself is stuck. That might depend on job specific logic.

I think there is a world where we do both separately.

But I think you're more interested in the first scenario - detecting jobs from crashed workers. I think you're particularly interested in this because you're setting timeout to None, so a job would hang _forever_ in this scenario.

Spitballing here...
In lieu of two separate solutions, maybe we can move this logic into the worker itself? For example, what if there was a "Job.automatic_heartbeats=True" parameter, and the Worker knew to set up that periodic task if the parameter is set to True? That feels a little cleaner to me... as we're really trying to fix some worker behavior here.

barakalon:
On second thought, I don't like the "automatic_heartbeats" idea.

If we're going to solve this, I think we need an explicit way to reclaim jobs from crashed workers.

examples/monitored.py
cofin: The idea behind this example is to show how the jobs fail when there is a missed heartbeat vs the job executing with the decorator.  It is a copy and paste of another example with a few modifications.

saq/job.py
cofin: Yep, you are correct.  It was a note re: why I made the modification, and it should have been removed.

saq/job.py
cofin: would something like `periodic_heartbeat` be more appropriate?

barakalon:
On third thought - 

I don't think this change is a good idea.

@cofin - worker crashes that fail to update jobs should be relatively rare. Have you considered adding a longer timeout in the event that they happen? Otherwise, I think it's best to manually instrument heartbeats within jobs. 

Detecting jobs that are stuck because a worker crashed and failed to update statuses is a legitimate use case, but it'll be more complex, and I don't think we should overload this heartbeating feature.

I'm closing this for now, but I'm happy to chat more about this.

tobymao:
@cofin we appreciate you taking the time to make these prs

cofin:
No problem.  I appreciate the quick code reviews.  Please let me know if there are areas where you specifically need contributions.

PR #37, opened 2022-08-16
fix: Replace exception logger function with logger.exception

sondrelg:
This PR tweaks the way errors are logged, to include more metadata.

I didn't spend a lot of time crafting the logger messages, so suggestions for better ones are very welcome 🙂 

## Motivation

We use [Sentry](sentry.io) for errors, and `saq` currently logs job exceptions in a way that's slightly too generic for Sentry. Since the plaintext dump of a traceback is logged without any extra information, all errors raised by `saq` are bundled together. Moreover, there is a size limit to log message length in Sentry, so we're not even receiving the full tracebacks (see screenshot at the bottom of the comment below).

image

Using `logger.exception` or `logger.error(..., exc_info=True)` (they're the same thing), we're able to get proper Sentry events, containing the complete traceback plus "breadcrumbs"

image

### DIFF
--- a/saq/worker.py
+++ b/saq/worker.py
@@ -15,12 +15,6 @@ from saq.utils import millis, now, seconds
 logger = logging.getLogger("saq")
 
 
-def get_and_log_exc():
-    error = traceback.format_exc()
-    logger.error(error)
-    return error
-
-
 class Worker:
     """
     Worker is used to process and monitor jobs.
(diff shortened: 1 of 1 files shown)


sondrelg:
To test, I created his task function to run. Once on the latest release, and once with the fork:

```python
def bad_function(ctx):
    raise ValueError('I fail now beep boop')
```

Here is the Sentry event produced with latest `saq` release:

image

And here is the Sentry event produced with the fork branch:

image

The second one could also use some extra information, but at least now we get the code that raised the exception in focus, rather than have it maybe be included in the traceback plaintext dump.

Another thing to mention is that Sentry's logging integration cuts off long log messages, so the plaintext dump (for us) gets chopped in half hehe

image

tobymao:
what is the result in the command line?

barakalon (approved):
Sweet.

I'm pretty sure the stream handler dumps the exception info properly.

Issue #39, opened 2022-09-09
Add type annotations


MrAwesomeRocks:
It would be really great if this project had PEP 484 type hints, for better editor support and integration with Mypy.

I'd be willing to help with this.

tobymao:
i’m against types annotations

MrAwesomeRocks:
How come?

tobymao:
python is not a typed language and the types don’t do anything, their fancy comments. i find them more of a hindrance than helpful. 

most type errors are due to having None. but  python doesn’t have a true option type with operators like map such as those found in scala.

peterschutt:
@tobymao I respect your position here, but would you mind if I created `saq-stubs` and publish to pypi so users can opt into using the stubs if they choose?

tobymao:
@peterschutt i've decided i cannot win this battle, i will accept a PR to saq directly if you add types and mypy to the whole repo.

peterschutt:
I'm sorry to be straw that broke the camel's back :laughing: - there really is no escaping it.

FWIW, I've been very comfortable using the library without annotations - the code is clean, concise and easy to follow. I just experience issue when trying to use a consistent set of tooling across apps.

I'll start with a stubs package, and once I have enough detail in the stubs to be useful, then we can see what it looks like in-line.

tobymao:
i'd much prefer just doing it in saq as opposed to doing stubs, let me know what i can do to help you make that happen

peterschutt:
OK great, thanks.

I'll probably end up trying a tool like [MonkeyType](https://github.com/instagram/MonkeyType) just to see if that can get us past the low-hanging fruit. I'll make a draft PR and we can go from there.

PR #43, opened 2022-10-21
Feat: Adds type annotations

peterschutt:
- Used MonkeyType and the test suite to generate  a set of annotations
- Added mypy and necessary stub dependencies to dev extras
- Adds types.py module for type definitions/aliases etc
- Adds py.typed
- Removes support for 3.7

Related to #39

### DIFF
--- a/saq/job.py
+++ b/saq/job.py
@@ -4,7 +4,7 @@ import dataclasses
 import enum
 import typing as t
 
-from saq.utils import now, seconds, uuid1, exponential_backoff
+from saq.utils import exponential_backoff, now, seconds, uuid1
 
 ABORT_ID_PREFIX = "saq:abort:"
 
--- a/saq/job.py
+++ b/saq/job.py
@@ -190,10 +190,8 @@ class Job:
         )
 
     def next_retry_delay(self) -> float:
-        if self.retry_backoff:
-            max_delay = self.retry_delay
-            if max_delay is True:
-                max_delay = None
+        if self.retry_backoff is not False:
+            max_delay = None if self.retry_backoff is True else self.retry_backoff
             return exponential_backoff(
                 attempts=self.attempts,
                 base_delay=self.retry_delay,
--- a/saq/job.py
+++ b/saq/job.py
@@ -10,6 +10,7 @@ ABORT_ID_PREFIX = "saq:abort:"
 
 if t.TYPE_CHECKING:
     from saq.queue import Queue
+    from saq.types import DurationKind, Function
 
 
 def get_default_job_key() -> str:
@@ -42,13 +43,13 @@ class CronJob:
     Remaining kwargs are pass through to Job
     """
 
-    function: t.Callable
+    function: Function
     cron: str
     unique: bool = True
-    timeout: t.Optional[int] = None
-    heartbeat: t.Optional[int] = None
-    retries: t.Optional[int] = None
-    ttl: t.Optional[int] = None
+    timeout: int | None = None
+    heartbeat: int | None = None
+    retries: int | None = None
+    ttl: int | None = None
 
 
 @dataclasses.dataclass
@@ -62,7 +63,7 @@ class Job:
         queue: the saq.Queue object associated with the job
         key: unique identifier of a job, defaults to uuid1, can be passed in to avoid duplicate jobs
         timeout: the maximum amount of time a job can run for in seconds, defaults to 10 (0 means disabled)
-        heartbeat: the maximum amount of time a job can survive without a heartebat in seconds, defaults to 0 (disabled)
+        heartbeat: the maximum amount of time a job can survive without a heartbeat in seconds, defaults to 0 (disabled)
             a heartbeat can be triggered manually within a job by calling await job.update()
         retries: the maximum number of attempts to retry a job, defaults to 1
         ttl: the maximum time in seconds to store information about a job including results, defaults to 600 (0 means indefinitely, -1 means disabled)
@@ -82,20 +83,20 @@ class Job:
         started: job started time epoch seconds
         touched: job touched/updated time epoch seconds
         results: payload containing the results, this is the return of the function provided, must be serializable, defaults to json
-        error: stack trace if an runtime error occurs
-        status: Status Enum, defaulst to Status.New
+        error: stack trace if a runtime error occurs
+        status: Status Enum, default to Status.New
     """
 
     function: str
-    kwargs: t.Optional[dict] = None
-    queue: t.Optional[Queue] = None
+    kwargs: dict | None = None
+    queue: Queue | None = None
     key: str = dataclasses.field(default_factory=get_default_job_key)
     timeout: int = 10
     heartbeat: int = 0
     retries: int = 1
     ttl: int = 600
     retry_delay: float = 0.0
-    retry_backoff: t.Union[bool, float] = False
+    retry_backoff: bool | float = False
     scheduled: int = 0
     progress: float = 0.0
     attempts: int = 0
@@ -104,7 +105,7 @@ class Job:
     started: int = 0
     touched: int = 0
     result: t.Any = None
-    error: t.Optional[str] = None
+    error: str | None = None
     status: Status = Status.NEW
     meta: dict = dataclasses.field(default_factory=dict)
 
@@ -146,7 +147,7 @@ class Job:
     def abort_id(self) -> str:
         return f"{ABORT_ID_PREFIX}{self.key}"
 
-    def to_dict(self) -> t.Dict[str, t.Any]:
+    def to_dict(self) -> dict[str, t.Any]:
         result = {}
         for field in dataclasses.fields(self):
             key = field.name
@@ -160,7 +161,7 @@ class Job:
             result[key] = value
         return result
 
-    def duration(self, kind: str) -> t.Optional[int]:
+    def duration(self, kind: DurationKind) -> int | None:
         """
         Returns the duration of the job given kind.
 
@@ -177,7 +178,7 @@ class Job:
             return self._duration(now(), self.started)
         raise ValueError(f"Unknown duration type: {kind}")
 
-    def _duration(self, a: int, b: int) -> t.Optional[int]:
+    def _duration(self, a: int, b: int) -> int | None:
         return a - b if a and b else None
 
     @property
@@ -200,12 +201,12 @@ class Job:
             )
         return self.retry_delay
 
-    async def enqueue(self, queue: t.Optional[Queue] = None) -> None:
+    async def enqueue(self, queue: Queue | None = None) -> None:
         """
         Enqueues the job to it's queue or a provided one.
 
         A job that already has a queue cannot be re-enqueued. Job uniqueness is determined by its id.
-        If a job has already been queued, it will update it's properties to match what is stored in the db.
+        If a job has already been queued, it will update its properties to match what is stored in the db.
         """
         queue = queue or self.get_queue()
         if not await queue.enqueue(self):
@@ -215,15 +216,17 @@ class Job:
         """Tries to abort the job."""
         await self.get_queue().abort(self, error, ttl=ttl)
 
-    async def finish(self, status: Status, *, result=None, error=None) -> None:
+    async def finish(
+        self, status: Status, *, result: t.Any = None, error: str | None = None
+    ) -> None:
         """Finishes the job with a Job.Status, result, and or error."""
         await self.get_queue().finish(self, status, result=result, error=error)
 
-    async def retry(self, error: t.Optional[str]) -> None:
-        """Retries the job by removing it from active and requeueing it."""
+    async def retry(self, error: str | None) -> None:
+        """Retries the job by removing it from active and enqueueing it again."""
         await self.get_queue().retry(self, error)
 
-    async def update(self, **kwargs) -> None:
+    async def update(self, **kwargs: t.Any) -> None:
         """
         Updates the stored job in redis.
 
@@ -233,7 +236,7 @@ class Job:
             setattr(self, k, v)
         await self.get_queue().update(self)
 
-    async def refresh(self, until_complete: t.Optional[float] = None) -> None:
+    async def refresh(self, until_complete: float | None = None) -> None:
         """
         Refresh the current job with the latest data from the db.
 
@@ -249,17 +252,16 @@ class Job:
 
         if until_complete is not None and not self.completed:
 
-            async def callback(_id, status):
-                if status in TERMINAL_STATUSES:
-                    return True
+            async def callback(_id: str, status: Status) -> bool:
+                return status in TERMINAL_STATUSES
 
             await self.get_queue().listen([self.key], callback, until_complete)
             await self.refresh()
 
     def replace(self, job: Job) -> None:
         """Replace current attributes with job attributes."""
-        for field in job.__dataclass_fields__:
-            setattr(self, field, getattr(job, field))
+        for field in dataclasses.fields(job):
+            setattr(self, field.name, getattr(job, field.name))
 
     def get_queue(self) -> Queue:
         if self.queue is None:
--- a/saq/job.py
+++ b/saq/job.py
@@ -88,7 +88,7 @@ class Job:
     """
 
     function: str
-    kwargs: dict | None = None
+    kwargs: dict[str, t.Any] | None = None
     queue: Queue | None = None
     key: str = dataclasses.field(default_factory=get_default_job_key)
     timeout: int = 10
@@ -107,7 +107,7 @@ class Job:
     result: t.Any = None
     error: str | None = None
     status: Status = Status.NEW
-    meta: dict = dataclasses.field(default_factory=dict)
+    meta: dict[t.Any, t.Any] = dataclasses.field(default_factory=dict)
 
     def __repr__(self) -> str:
         kwargs = ", ".join(
--- a/saq/job.py
+++ b/saq/job.py
@@ -88,7 +88,7 @@ class Job:
 
     function: str
     kwargs: t.Optional[dict] = None
-    queue: t.Optional["Queue"] = None
+    queue: t.Optional[Queue] = None
     key: str = dataclasses.field(default_factory=get_default_job_key)
     timeout: int = 10
     heartbeat: int = 0
@@ -259,7 +259,7 @@ class Job:
             await self.queue.listen([self.key], callback, until_complete)
             await self.refresh()
 
-    def replace(self, job: "Job") -> None:
+    def replace(self, job: Job) -> None:
         """Replace current attributes with job attributes."""
         for field in job.__dataclass_fields__:
             setattr(self, field, getattr(job, field))
--- a/saq/job.py
+++ b/saq/job.py
@@ -1,13 +1,18 @@
+from __future__ import annotations
+
 import dataclasses
 import enum
 import typing
 
-from saq.utils import now, seconds, uuid1, exponential_backoff
+from saq.utils import exponential_backoff, now, seconds, uuid1
 
 ABORT_ID_PREFIX = "saq:abort:"
 
+if typing.TYPE_CHECKING:
+    from saq.queue import Queue
 
-def get_default_job_key():
+
+def get_default_job_key() -> str:
     return uuid1()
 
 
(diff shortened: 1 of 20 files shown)


peterschutt:
@tobymao you should be able to run `python -m mypy saq/queue.py` without error - I've focused my attention there so far and would like to give you the chance to weigh in before I go any further.

I'm running with default config to start with, we can ramp up the strictness settings of mypy once I'm past the initial batch of errors.

Found an issue with the redis stubs in typeshed: https://github.com/python/typeshed/issues/8960

saq/job.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import dataclasses
 import enum
 import typing
```
tobymao, line 5: let's refactor everywhere that uses typing to the convention

import typing as t

saq/job.py
```diff
@@ -254,7 +259,7 @@ async def callback(_id, status):
             await self.queue.listen([self.key], callback, until_complete)
             await self.refresh()
 
-    def replace(self, job):
+    def replace(self, job: "Job") -> None:
```
tobymao, line 262: all types should be real types, not strings, let's just do Job here

saq/queue.py
```diff
@@ -20,20 +23,36 @@
     uuid1,
 )
 
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+    from typing import Any, Callable, Dict, List, Optional, Set, Type, Union
+    from redis.asyncio.client import Redis
+    from redis.commands.core import AsyncScript
+    from .types import (
```
barakalon, line 31: Should we avoid relative imports?

`from saq.types`

barakalon (approved):
yay!

Let's add mypy to `run_checks.sh`

saq/types.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, TypedDict
+
+if TYPE_CHECKING:
+    from typing import Any, Awaitable, Callable, Dict, Tuple, Union
+    from typing_extensions import TypeAlias
+    from .job import Job
```
tobymao, line 8: remove relative imports

saq/types.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, TypedDict
```
tobymao, line 3:
```suggestion
import typing as t
```

saq/types.py
```diff
@@ -0,0 +1,30 @@
...
+
+if TYPE_CHECKING:
+    from typing import Any, Awaitable, Callable, Dict, Tuple, Union
+    from typing_extensions import TypeAlias
+    from .job import Job
+
+BeforeEnqueueType: TypeAlias = "Callable[[Job], Awaitable[Any]]"
```
tobymao, line 10: don't use strings

peterschutt:
Thanks for the reviews! I'll keep working at it over the weekend.

> Let's add mypy to run_checks.sh

172 errors still when I run on saq and tests :) will def. add it in when I get a clean run.

tobymao:
let me know when is hould take another look

peterschutt:
A good chunk of the mypy errors revolve around attribute access on `Job.queue` as it might be `None`:

```
saq/job.py:117: error: Item "None" of "Optional[Queue]" has no attribute "name"  [union-attr]
saq/job.py:139: error: Item "None" of "Optional[Queue]" has no attribute "job_id"  [union-attr]
saq/job.py:217: error: Item "None" of "Optional[Queue]" has no attribute "abort"  [union-attr]
saq/job.py:221: error: Item "None" of "Optional[Queue]" has no attribute "finish"  [union-attr]
saq/job.py:225: error: Item "None" of "Optional[Queue]" has no attribute "retry"  [union-attr]
saq/job.py:235: error: Item "None" of "Optional[Queue]" has no attribute "update"  [union-attr]
saq/job.py:244: error: Item "None" of "Optional[Queue]" has no attribute "job"  [union-attr]
saq/job.py:257: error: Item "None" of "Optional[Queue]" has no attribute "listen"  [union-attr]
```
These are currently `AttributeError`s:

```python-traceback
  File "/home/peter/.pyenv/versions/3.10.8/lib/python3.10/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "/home/peter/PycharmProjects/saq/saq/job.py", line 139, in id
    return self.queue.job_id(self.key)
AttributeError: 'NoneType' object has no attribute 'job_id'
```
One idea is to add an access method to centralise the exception logic and type narrowing, such as:

```python
class Job:
    ...
    def get_queue(self) -> Queue:
        if self.queue is None:
            raise TypeError("`Job` must be associated with a `Queue` before this operation can proceed")
        return self.queue

    @property
    def id(self) -> str:
        return self.get_queue().job_id(self.key)
```

It could also be a property and/or memoized but I doubt we'd be looking at too much overhead with the plain method.

Any opinion here @tobymao @barakalon?

tobymao:
> A good chunk of the mypy errors revolve around attribute access on `Job.queue` as it might be `None`:
> 
> ```
> saq/job.py:117: error: Item "None" of "Optional[Queue]" has no attribute "name"  [union-attr]
> saq/job.py:139: error: Item "None" of "Optional[Queue]" has no attribute "job_id"  [union-attr]
> saq/job.py:217: error: Item "None" of "Optional[Queue]" has no attribute "abort"  [union-attr]
> saq/job.py:221: error: Item "None" of "Optional[Queue]" has no attribute "finish"  [union-attr]
> saq/job.py:225: error: Item "None" of "Optional[Queue]" has no attribute "retry"  [union-attr]
> saq/job.py:235: error: Item "None" of "Optional[Queue]" has no attribute "update"  [union-attr]
> saq/job.py:244: error: Item "None" of "Optional[Queue]" has no attribute "job"  [union-attr]
> saq/job.py:257: error: Item "None" of "Optional[Queue]" has no attribute "listen"  [union-attr]
> ```
> 
> These are currently `AttributeError`s:
> 
> ```
> Traceback (most recent call last):
>   File "/home/peter/.pyenv/versions/3.10.8/lib/python3.10/code.py", line 90, in runcode
>     exec(code, self.locals)
>   File "<input>", line 1, in <module>
>   File "/home/peter/PycharmProjects/saq/saq/job.py", line 139, in id
>     return self.queue.job_id(self.key)
> AttributeError: 'NoneType' object has no attribute 'job_id'
> ```
> 
> One idea is to add an access method to centralise the exception logic and type narrowing, such as:
> 
> ```python
> class Job:
>     ...
>     def get_queue(self) -> Queue:
>         if self.queue is None:
>             raise TypeError("`Job` must be associated with a `Queue` before this operation can proceed")
>         return self.queue
> 
>     @property
>     def id(self) -> str:
>         return self.get_queue().job_id(self.key)
> ```
> 
> It could also be a property and/or memoized but I doubt we'd be looking at too much overhead with the plain method.
> 
> Any opinion here @tobymao @barakalon?

that’s fine

saq/job.py
```diff
@@ -233,7 +236,7 @@ async def update(self, **kwargs) -> None:
             setattr(self, k, v)
         await self.get_queue().update(self)
 
-    async def refresh(self, until_complete: t.Optional[float] = None) -> None:
+    async def refresh(self, until_complete: float | None = None) -> None:
```
tobymao, line 239: is this syntax legal in 3.7?

peterschutt: I was just about to write an explanation.

The `from __future__ import annotations` means that all annotations are strings at runtime, so no runtime errors irrespective of version, and type checkers are able to use the current syntax irrespective of the version.

I've been running the tests locally on 3.8 without issue and that syntax was introduced in 3.10. Similarly, the builtin subscription syntax, e.g., `list[...]` is 3.9+ but the future annotations mean we don't need to worry about the runtime errors.

saq/types.py
```diff
@@ -0,0 +1,43 @@
...
+    retried: int
+    aborted: int
+    uptime: int
+
+
+class TimersDict(TypedDict, total=False):
+    schedule: int
+    stats: int
+    sweep: int
+    abort: int
```
peterschutt, line 43: @tobymao Note in this module that I've used the typing versions of generics as it would be feasible that a user would import something from here to annotate their own code, and we can't be sure they'll use it with `from __future__ import annotations`, so needs to be compatible with 3.7 runtime.

This is also why `"Job"` and `"Status"` are forward refs here, as if this module is imported outside of a `if TYPE_CHECKING` block, their reference would otherwise fail.

saq/job.py
peterschutt: Of course, if there is an issue I'm missing here, or you'd prefer to stay syntactically valid with the min version irrespective of the future annotations, I'll be happy to roll them back. Just that the current syntax is a lot nicer than what was available in 3.7.

saq/types.py
```diff
@@ -0,0 +1,43 @@
...
+
+if t.TYPE_CHECKING:
+    from saq.job import Job, Status
+
+BeforeEnqueueType = t.Callable[["Job"], t.Awaitable[t.Any]]
+Context = t.Dict[str, t.Any]
+CountKind = Literal["queued", "active", "incomplete"]
+DumpType = t.Callable[[t.Dict], str]
+DurationKind = Literal["process", "start", "total", "running"]
+Function = t.Callable
```
peterschutt, line 15: This is just a stub for a better type.

I tried different configurations with `Protocol` and `__call__` and `@overload` to try and satisfy requirement that these can be both sync and async, accept a single pos arg, either accept kwargs, or not, as well as having the `__qualname__` attribute, but couldn't find a solution that worked.

saq/worker.py
```diff
@@ -1,16 +1,28 @@
...
 
 from croniter import croniter
 
 from saq.job import Status
 from saq.queue import Queue
 from saq.utils import millis, now, seconds
 
+if t.TYPE_CHECKING:
+    from asyncio import Task
+    from collections.abc import Callable, Collection, Coroutine
```
peterschutt, line 19: Are you happy for individual imports for `collections.abc`? Or prefer `from collections import abc` and `abc.Callable[...]` etc?

tobymao: it’s ok

saq/types.py
tobymao: can you double check everything works fine in 3.7?

peterschutt: yes I will verify

saq/job.py
barakalon: I think mypy will automatically add an Optional type when the default value is None? I could have that wrong, though.

saq/types.py
peterschutt: `IsolatedAsyncioTestCase` not available in 3.7. I tried to do a compat with [aiounittest](https://aiounittest.readthedocs.io/en/latest/info.html) for 3.7 (`import aiounittest.AsyncTestCase as `IsolatedAsyncioTestCase`) but there were a stack of errors. 

Perhaps a simple integration test script can be added to CI for 3.7 only?

3.7 EOL is 27th June 2023.

barakalon: I've manually ported IsolatedAsyncioTestCase before. I can help with that.

@tobymao - is there some specific 3.7 use case you have in mind?

tobymao: i guess we can leave saq at 3.8+, in my mind everything i support is 3.7+ because of airflow, but i guess there's no reason to mix airflow and saq

barakalon: Then maybe we ignore 3.7?

tobymao: yes

peterschutt: that would certainly simplify things

saq/job.py
peterschutt: You are correct, but reliance on it is discouraged: https://github.com/python/peps/pull/689 and https://github.com/python/mypy/issues/9091

saq/types.py
peterschutt: 3.7 removed and added 3.11

saq/types.py
peterschutt: Going 3.8+ might help with this as positional only parameters were introduced in 3.8.

saq/types.py
peterschutt: Note we still need to use the typing module generics here unless we go back to:

```python
LoadType: t.TypeAlias = "Callable[[bytes | str], t.Any]"
```
Six to one, half a dozen to the other in my mind, so I'll leave as is unless you'd prefer otherwise.

peterschutt:
I just added a commit that addresses issues with public interface types highlighted by the Pyright `--verifytypes` tool. For context, this was the output of the first run of the tool before any fixes.

Most of the issues were from use of generic types without specifying type arguments.

One interesting case to highlight was this:

```
saq.queue.Queue.__init__
  /home/peter/PycharmProjects/saq/saq/queue.py:72:9 - error: Type of parameter "redis" is partially unknown
    Parameter type is "Redis[Unknown]"
      Type argument 1 for class "Redis" has unknown type
```

`Redis` can be configured with `decode_responses=True` so that it returns `str` from commands, which would mean code like this would fail:

```python
        for key in await self.redis.zrangebyscore(self._stats, now(), "inf"):
            key_str = key.decode("utf-8")
```

Therefore `Queue` specifically expects `decode_responses=False` which is `Redis[bytes]`.

On the negative, it raised a few like this:

```
saq.worker.logger
  /home/peter/PycharmProjects/saq/saq/worker.py:27:1 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "Logger"

```

Meaning it wants:

```python
logger: logging.Logger = logging.getLogger("saq")
```

Pretty gross, and I'm not sure how that could be inferred to be anything different by other type checkers.

Full output:
```
(.venv) peter@peter-Inspiron-15-5510:~/test-saq$ pyright --verifytypes saq --ignoreexternal
Module name: "saq"
Package directory: "/home/peter/PycharmProjects/saq/saq"
Module directory: "/home/peter/PycharmProjects/saq/saq"
Path of py.typed file: "/home/peter/PycharmProjects/saq/saq/py.typed"

Public modules: 8
   saq
   saq.__main__
   saq.job
   saq.queue
   saq.types
   saq.utils
   saq.web
   saq.worker

Symbols used in public interface:
saq.job.Job.kwargs
  /home/peter/PycharmProjects/saq/saq/job.py:91:5 - error: Type argument 1 for class "dict" has unknown type
  /home/peter/PycharmProjects/saq/saq/job.py:91:5 - error: Type argument 2 for class "dict" has unknown type
saq.queue.Queue.__init__
  /home/peter/PycharmProjects/saq/saq/queue.py:72:9 - error: Type of parameter "redis" is partially unknown
    Parameter type is "Redis[Unknown]"
      Type argument 1 for class "Redis" has unknown type
saq.queue.Queue.redis
  /home/peter/PycharmProjects/saq/saq/queue.py:80:14 - error: Type argument 1 for class "Redis" has unknown type
saq.queue.Queue.uuid
  /home/peter/PycharmProjects/saq/saq/queue.py:82:14 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "str"
saq.queue.Queue.started
  /home/peter/PycharmProjects/saq/saq/queue.py:83:14 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "int"
saq.job.Job.meta
  /home/peter/PycharmProjects/saq/saq/job.py:110:5 - error: Type argument 1 for class "dict" has unknown type
  /home/peter/PycharmProjects/saq/saq/job.py:110:5 - error: Type argument 2 for class "dict" has unknown type
saq.job.Job.enqueue
  /home/peter/PycharmProjects/saq/saq/job.py:204:15 - error: Type of parameter "queue" is partially unknown
    Parameter type is "Queue | None"
saq.job.Job.get_queue
  /home/peter/PycharmProjects/saq/saq/job.py:266:9 - error: Return type is partially unknown
    Return type is "Queue"
saq.queue.logger
  /home/peter/PycharmProjects/saq/saq/queue.py:38:1 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "Logger"
saq.queue.JobError.__init__
  /home/peter/PycharmProjects/saq/saq/queue.py:44:9 - error: Type of parameter "job" is partially unknown
    Parameter type is "Job"
saq.types.QueueInfo.workers
  /home/peter/PycharmProjects/saq/saq/types.py:21:5 - error: Type argument 1 for class "dict" has unknown type
  /home/peter/PycharmProjects/saq/saq/types.py:21:5 - error: Type argument 2 for class "dict" has unknown type
saq.types.QueueInfo.jobs
  /home/peter/PycharmProjects/saq/saq/types.py:26:5 - error: Type argument 1 for class "list" has partially unknown type
    Type is dict[Unknown, Unknown]
saq.web.static
  /home/peter/PycharmProjects/saq/saq/web.py:23:1 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "str"
saq.worker.logger
  /home/peter/PycharmProjects/saq/saq/worker.py:27:1 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "Logger"
saq.worker.Worker.__init__
  /home/peter/PycharmProjects/saq/saq/worker.py:50:9 - error: Type of parameter "queue" is partially unknown
    Parameter type is "Queue"
  /home/peter/PycharmProjects/saq/saq/worker.py:50:9 - error: Type of parameter "functions" is partially unknown
    Parameter type is "Collection[((...) -> Unknown) | tuple[str, (...) -> Unknown]]"
      Type argument 1 for class "Collection" has partially unknown type
saq.worker.Worker.upkeep
  /home/peter/PycharmProjects/saq/saq/worker.py:161:15 - error: Return type is partially unknown
    Return type is "Coroutine[Any, Any, list[Task[Unknown]]]"
      Type argument 3 for class "Coroutine" has partially unknown type
saq.worker.Worker.event
  /home/peter/PycharmProjects/saq/saq/worker.py:111:18 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "Event"
saq.worker.Worker.functions
  /home/peter/PycharmProjects/saq/saq/worker.py:80:14 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "dict[Unknown, Unknown]"
  /home/peter/PycharmProjects/saq/saq/worker.py:80:14 - error: Type argument 1 for class "dict" has unknown type
  /home/peter/PycharmProjects/saq/saq/worker.py:80:14 - error: Type argument 2 for class "dict" has unknown type
saq.worker.Worker.cron_jobs
  /home/peter/PycharmProjects/saq/saq/worker.py:81:14 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "Collection[CronJob] | list[CronJob]"
saq.worker.Worker.tasks
  /home/peter/PycharmProjects/saq/saq/worker.py:83:14 - error: Type argument 1 for class "set" has partially unknown type
    Type is Task[Unknown]
saq.worker.Worker.job_task_contexts
  /home/peter/PycharmProjects/saq/saq/worker.py:84:14 - error: Type argument 2 for class "dict" has partially unknown type
    Type is dict[Unknown, Unknown]
saq.worker.async_check_health
  /home/peter/PycharmProjects/saq/saq/worker.py:332:11 - error: Type of parameter "queue" is partially unknown
    Parameter type is "Queue"

Symbols exported by "saq": 203
  With known type: 173
  With ambiguous type: 7
  With unknown type: 23
    (Ignoring unknown types imported from other packages)
  Functions without docstring: 50
  Functions without default param: 0
  Classes without docstring: 5

Other symbols referenced but not exported by "saq": 0
  With known type: 0
  With ambiguous type: 0
  With unknown type: 0

Type completeness score: 85.2%

Completed in 0.874sec
```

tobymao:
seems reasonable to me, is this ready to go?

peterschutt:
> seems reasonable to me, is this ready to go?

Close, I'm experimenting with it downstream. The top-level import aren't being recognized by mypy, e.g.,:

```python
import saq

q = saq.Queue()   # Module has no attribute "Queue"  [attr-defined]
```

Might need an `__all__` but I'm looking into it.

peterschutt:
> > seems reasonable to me, is this ready to go?
> 
> Close, I'm experimenting with it downstream. The top-level import aren't being recognized by mypy, e.g.,:
> 
> ```python
> import saq
> 
> q = saq.Queue()   # Module has no attribute "Queue"  [attr-defined]
> ```
> 
> Might need an `__all__` but I'm looking into it.

That was it ([ref](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport)).

Should be good to go now.

tobymao:
@peterschutt thanks again for this, great work!

peterschutt:
> @peterschutt thanks again for this, great work!

No worries! Feel free to ping me if there are any issues.

PR #64, opened 2023-04-09
Require redis >=4.2,<4.6

### LINKED ISSUE #63: Support redis 4.5.x?
1. There's a race condition bug with redis <=4.4.3 and <=4.5.3, which was addressed in the next bugfix versions.
2. I'm trying to install litestar 2.0.0alpha3, which depends on redis >=4.4.4 and >=4.5.4 (yeah they explicitly require so to avoid that race condition bug).

lqmanh:
Resolve #63

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,6 @@
 # pylint: disable=consider-using-with
 from setuptools import setup
 
-
 version = (
     open("saq/__init__.py", encoding="utf8")
     .read()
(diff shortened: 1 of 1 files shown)


tobymao:
this is not correct because redis broke cancel catching in 4.4 look at 

https://github.com/redis/redis-py/pull/2104

tobymao:
this is still open https://github.com/redis/redis-py/issues/2511

lqmanh:
Okay this seems really bad. Guess I have to find some workaround then :(((
Thanks for your quick response 👍🏼

kartzke:
Looks like this PR may fix the issue:  https://github.com/redis/redis-py/pull/2695

It's been hanging for 3 months.  There's been some dialog in the last few days, but it needs the attention of a redis-py maintainer.

Issue #67, opened 2023-05-15
Redis requirement ranges


sondrelg:
Hi @tobymao, sorry for the issue spam.

Would you mind unpinning the redis requirement upper-bound? I see the [original issue](https://github.com/redis/redis-py/pull/2104) is resolved, and I need to be able to run 4.4+ to fix another ][issue](https://github.com/redis/redis-py/issues/2449) I'm having, migrating from a standalone to cluster redis setup :bow: 

Specifically this line: https://github.com/tobymao/saq/blob/master/setup.py#L32

tobymao:
do you know how to do version bounds?

PR #68, opened 2023-05-15
Unpin redis requirement

sondrelg:
Relates to #67 

Note: I haven't deeply looked into the original reason for why it was pinned; just saw it had been resolved a while back.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
         saq=saq.__main__:main
     """,
     install_requires=[
-        "redis>=4.2,<4.4",
+        "redis>=4.2",
         "croniter>=0.3.18",
     ],
     extras_require={


sondrelg:
Looks like there are other things preventing us from using saq on a Redis cluster. Saw a lot of 

```
redis.exceptions.RedisClusterException: transaction is deprecated in cluster mode
``` 

and 
```
redis.exceptions.RedisClusterException: EVALSHA - all keys must map to the same key slot
```

but I suppose the redis dependency might still be possible to unpin :relaxed:

tobymao:
i'll fix it on my end

sondrelg:
Why not make it `<5`?

sondrelg:
Ah cool, that's what you ended up doing. Sweet :relaxed:

tobymao:
we’ll you can use 4.5.4 either…

PR #72, opened 2023-06-24
Add option to have SAQ use fifo instead of filo

ZaSkittles:
Currently SAQ uses first in last out by default. 
This can create problems if lots of new jobs are being enqueued. Leaving this as the current default behavior to not surprise anyone, but creating an option to change it.


tobymao:
let’s make it default

tobymao:
actually get rid of the option, make this the only way

ZaSkittles:
sg just updated

PR #79, opened 2023-08-30
Add mountable Starlette web app

grigi:
For use in anything that is derived from Starlette.

For Starlette:
```py
from saq.web.starlette import saq_web
from starlette.routing import Mount

routes = [
    ...
    Mount("/monitor", saq_web("/monitor", queues=all_the_queues))
]
```

For FastAPI:
```py
from fastapi import FastAPI
from saq.web.starlette import saq_web

app = FastAPI()

app.mount("/monitor", saq_web("/monitor", queues=all_the_queues))
```

Also:
* I also did some minor refactoring to put the existing aiohttp app in `saq/web/aiohttp.py` and made the saq tools/test import the correct instance.
* Web interface has a `root_path` that is explicitly managed on server-side and JS side. Pass `root_path` as a variable with the html template instead of modifying the JS.
* Starlette unfortunately doesn't allow having any types used not imported, as it uses it internally. Therefore I had to bypass `typing.TYPE_CHECKING` for anything that's exported via the web interface.
* Changed `Docs` link to point to RTD docs.
* Added docs

Outstanding:
* Tests


tobymao:
@grigi do you want to become a maintainer?

tobymao:
what’s the use case of this change? saq is meant to run alone in aiotthp

grigi:
> @grigi do you want to become a maintainer?

Um... maybe? But not right now.

> what’s the use case of this change? saq is meant to run alone in aiotthp

We want to be able to mount the web UI in our internal http api, which is implemented in FastAPI. That will simplify network routing.

tobymao (approved):

PR #84, opened 2023-08-31
Added package saq.web to build

grigi:
Hi, thanks for releasing a new version, but it seems that the saq.web pacakges are not included. (my bad, I didn't verify that setup.py didn't use `find_packages()`)

I also updated the CHANGELOG assuming the next release should be called v0.11.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,16 @@
 # Changelog
 
-## Next
-- `saq` command-line client includes current path to simplify development
+## v0.11
+
+### v0.11.2
+- Added package `saq.web` to build
+
+### v0.11.1
 - Added embeddable `saq.web.starlette`
+
+### v0.11.0
+- `saq` command-line client includes current path to simplify development
+- Fetch outstanding jobs from incomplete namespace instead of queued.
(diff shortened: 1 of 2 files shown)


tobymao:
the changelog should not be manual, i will never update it, you need to add an automatic changelog tool like semantic commits or something

grigi:
> the changelog should not be manual, i will never update it, you need to add an automatic changelog tool like semantic commits or something

Noted, I'll look for something that can automate it or get rid of it.

Otherwise I noticed *another* packaging mistake :sob: 
```
RuntimeError: Directory '/home/grigi/work/magneto/.pyenv/lib/python3.11/site-packages/saq/web/static' does not exist
```

I'm making dinner now and dealing with the kids, so I'll only get to it tomorrow morning. Sorry

tobymao:
219e702

can you check if this does it when you have time?

tobymao:
https://github.com/tobymao/saq/commit/7e6e7739707a373ca8fff1ebed311849189b5ddf

grigi:
> [7e6e773](https://github.com/tobymao/saq/commit/7e6e7739707a373ca8fff1ebed311849189b5ddf)

Yay! It's working perfectly now!
Thank you sir :bow:

PR #85, opened 2023-08-31
Autogenerate changelog off of git commits and tags

grigi:
I tried 3 different libraries, and this one generates the second-best docs, but it doesn't require you to manually manage github releases. Unfortunately it's a little hacky :no_mouth: 

I also only _found_ 3 libraries in my search. There are probably more :shrug: 

It generates a changelog that looks like this:
image


tobymao (approved):

tobymao:
right now i release by putting a tag, can i continue to do that?

grigi:
Yes, I made it work with the minimum friction based on what you're doing.
Just tagging in the `vX.Y` or `vX.Y.Z` format will mark a version.

That's why the one that generated the nicest docs lost, as it would require a workflow change.

I mean, this is decent enough for no extra effort: https://saq-py.readthedocs.io/en/latest/changelog.html

PR #90, opened 2023-09-11
Logging verbosity balancing.

grigi:
I prefer a modicum of logging to simplify diagnostics, but not too much.

* Default loglevel of `ERROR` was too restrictive as warnings won't get reported. I changed the default loglevel to include `WARNING`.
* The Worker would not log when it started up, so I added an `INFO` for that, it includes logging the `Queue` config being used for easy identification of the worker.
* The Worker now logs the configured functions as `DEBUG`.
* The `INFO` loglevel would dump the entire payload and extra data for each job which would fill up the logs with more data, this was quite a jump from no logs at all for job processing. I split it up so that if loglevel is `INFO` it will emit an abridged version of the Job, and if loglevel is `DEBUG` it will emit the full version. (I'm not 100% happy about this, but needed to just log less of the same data for INFO by default)

### DIFF
--- a/saq/job.py
+++ b/saq/job.py
@@ -129,7 +129,20 @@ class Job:
     status: Status = Status.NEW
     meta: dict[t.Any, t.Any] = dataclasses.field(default_factory=dict)
 
-    def __repr__(self) -> str:
+    _EXCLUDE_NON_FULL = {
+        "kwargs",
+        "scheduled",
+        "progress",
+        "total_ms",
+        "result",
+        "error",
+        "status",
+        "meta",
+    }
+
+    def info(self, full: bool = False) -> str:
+        # Using an exclusion list preserves order for kwargs below
+        excluded = set() if full else self._EXCLUDE_NON_FULL
         kwargs = ", ".join(
             f"{k}={v}"
             for k, v in {
@@ -148,9 +161,12 @@ class Job:
                 "status": self.status,
                 "meta": self.meta,
             }.items()
-            if v is not None
+            if v is not None and k not in excluded
         )
         return f"Job<{kwargs}>"
+
+    def __repr__(self) -> str:
+        return self.info(True)
 
     def __hash__(self) -> int:
         return hash(self.key)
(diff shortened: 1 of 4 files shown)


saq/job.py
```diff
@@ -129,6 +129,23 @@ class Job:
     status: Status = Status.NEW
     meta: dict[t.Any, t.Any] = dataclasses.field(default_factory=dict)
 
+    def __str__(self) -> str:
```
tobymao, line 132: why not have this just be the same as repr?

saq/queue.py
```diff
@@ -529,7 +541,10 @@ async def enqueue(self, job_or_func: str | Job, **kwargs: t.Any) -> Job | None:
             ):
                 return None
 
-        logger.info("Enqueuing %s", job)
+        logger.info(
+            "Enqueuing %s",
+            repr(job) if logger.isEnabledFor(logging.DEBUG) else str(job),
```
tobymao, line 546: maybe make this a helper function because it's very repetitive. i don't like how repr and str are different, perhaps it could be an argument to the helper function for simple or verbose logging

grigi: That's a good point. Unfortunately can't add arguments to repr or str, but a helper function, sure.

saq/job.py
grigi: __repr__ now just calls info(True)

saq/job.py
```diff
@@ -129,7 +129,22 @@ class Job:
     status: Status = Status.NEW
     meta: dict[t.Any, t.Any] = dataclasses.field(default_factory=dict)
 
-    def __repr__(self) -> str:
+    def info(self, full: bool = False) -> str:
+        # Using an exclusion list preserves order for kwargs below
+        excluded = (
+            set()
+            if full
+            else {
```
tobymao, line 137: can you make this a constant?

tobymao (approved):

saq/queue.py
grigi: I created an `info()` method that takes an optional boolean for `full`. This now simplifies to `job.info(logger.isEnabledFor(logging.DEBUG))` which is still more verbose than I would like, but we don't know the status of the logger at that point unless we pass it in.
And I decided that it was better to keep the `info()` function's behaviour independent of global state.

saq/job.py
grigi: Updated

PR #92, opened 2023-09-21
Use provided json encoder/decoder for stats & info queries

fadedDexofan:


tobymao:
i don't think this works, json.loads expects a string but you're going get bytes

fadedDexofan:
> i don't think this works, json.loads expects a string but you're going get bytes

From Python 3.6 arg `s` in json.loads can also be `bytes` and `bytearray`

tobymao:
can you fix the build?

fadedDexofan:
Working on it

PR #100, opened 2023-11-14
fix: Queue.listen connection leak

barakalon:
Queue.listen has a connection leak.

This fixes it by moving to a singleton pubsub instance that routes message in-process.

### DIFF
--- a/saq/queue.py
+++ b/saq/queue.py
@@ -8,6 +8,7 @@ import json
 import logging
 import time
 import typing as t
+from collections import defaultdict
 from contextlib import asynccontextmanager
 
 from redis import asyncio as aioredis
@@ -24,7 +25,7 @@ from saq.utils import millis, now, seconds, uuid1
 if t.TYPE_CHECKING:
     from collections.abc import AsyncIterator, Iterable, Sequence
 
-    from redis.asyncio.client import Redis
+    from redis.asyncio.client import Redis, PubSub
     from redis.commands.core import AsyncScript
 
     from saq.types import (
@@ -107,6 +108,9 @@ class Queue:
         self._load = load or json.loads
         self._before_enqueues: dict[int, BeforeEnqueueType] = {}
         self._op_sem = asyncio.Semaphore(max_concurrent_ops)
+        self._pubsub = PubSubMultiplexer(
+            redis.pubsub(), prefix=f"{ID_PREFIX}{self.name}"
+        )
 
     def __repr__(self) -> str:
         return f"{self.__class__.__name__}<redis={self.redis}, name='{self.name}'>"
@@ -140,6 +144,7 @@ class Queue:
         return Job(**job_dict, queue=self)
 
     async def disconnect(self) -> None:
+        await self._pubsub.close()
         await self.redis.close()
         await self.redis.connection_pool.disconnect()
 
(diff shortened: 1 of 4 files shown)


saq/queue.py
```diff
@@ -687,3 +696,70 @@ async def track_child(job: Job) -> None:
...
+        while True:
+            try:
+                message = await self.pubsub.get_message(timeout=None)  # type: ignore
+                if message and message["type"] == "pmessage":
+                    for queue in self._subscriptions[message["channel"]]:
+                        queue.put_nowait(message)
+            except asyncio.CancelledError:
+                return
+            except Exception:
+                logger.exception("Failed to consume message")
```
tobymao, line 733: show the message and stack trace maybe?

saq/queue.py
```diff
@@ -687,3 +696,70 @@ async def track_child(job: Job) -> None:
...
+
+    async def start(self) -> None:
+        if not self._daemon_task:
+            await self.pubsub.psubscribe(f"{self.prefix}*")
```
tobymao, line 737: f string is not necessary

saq/queue.py
```diff
@@ -687,3 +699,67 @@ async def track_child(job: Job) -> None:
...
+        # Use bytes so the daemon needs to do less work
+        bchannels = [c.encode() for c in channels]
+
+        self._queues[queue] = set(bchannels)
+        for channel in bchannels:
+            self._subscriptions[channel].add(queue)
+        return queue
+
+    async def unsubscribe(self, queue: Q) -> None:
+        channels = self._queues.pop(queue, None)
```
tobymao, line 762: for channel in self.queue.pop(queue, []):

saq/queue.py
```diff
@@ -328,31 +336,35 @@ async def listen(
             callback: callback function, if it returns truthy, break
             timeout: if timeout is truthy, wait for timeout seconds
         """
-        pubsub = self.redis.pubsub()
+        if not job_keys:
```
tobymao, line 339: why do you need this check? you can't check if job_keys is empty because it's an iterable

saq/queue.py
barakalon: logger.exception does that by default

saq/queue.py
barakalon: no?

saq/queue.py
barakalon: whoops, good catch.

Doesn't really make sense to PubSubMultiplexer.subscribe().
I'll just move the check to job_ids.

saq/queue.py
tobymao: sorry, missed the *

tobymao (approved):

tobymao:
@barakalon just for curiosity, can you run the benchmarks pre and post? you can adjust the benches to leverage this new stuff

barakalon:
Struggling a bit with the `rq` benchmarks, since the newer version logs a bunch of crap.
But there's no noticeable change in saq numbers.

tobymao:
> Struggling a bit with the `rq` benchmarks, since the newer version logs a bunch of crap. But there's no noticeable change in saq numbers.

don't worry about RQ. ok let's merge it

PR #102, opened 2023-11-29
use redis.aclose instead of close

tiejunhu:
redis.close is deprecated since version 5.0.0

### DIFF
--- a/saq/queue.py
+++ b/saq/queue.py
@@ -148,7 +148,7 @@ class Queue:
 
     async def disconnect(self) -> None:
         await self._pubsub.close()
-        await self.redis.close()
+        await self.redis.aclose()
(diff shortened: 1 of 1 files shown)


barakalon:
saq support redis>=4.2,<6.0
redis 4.2 doesn't have `aclose`: https://github.com/redis/redis-py/blob/v4.2.0/redis/asyncio/client.py
So we'd have to drop support for some redis versions. I'm not sure that's worth it?

tobymao:
can you adjust this pr to check if aclose exists?

tobymao:
closing for now, happy to receive a pr with redis 4 compatibility

tiejunhu:
ok. if redis 4 should be supported. let's keep the 'close' until it breaks on newer redis library.

PR #107, opened 2023-12-13
Update queue.py to support redis >= 5.* aclose method.

maxyme-perceiv:
Use redis's aclose() method if available otherwise fallback to close().
Note: Follow-up to [PR #102](https://github.com/tobymao/saq/pull/102)


tobymao:
thanks!

PR #112, opened 2024-02-19
Support settings factories

pierec:
Allow the settings objects imported during `saq.worker.start()` to be callable. This makes it possible to eliminate import-time side effects, especially if creating the settings objects is more involved.

In my case, I'm leveraging multiple queues and using a dependency injection framework to wire the `saq.Queue`s a certain way. While I could sequester the settings objects to their separate modules, I find it more convenient to keep them together.

Let me know if this idea is OK and thank you for this neat little framework!

### DIFF
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -2,14 +2,20 @@ from __future__ import annotations
 
 import asyncio
 import contextvars
+import contextlib
 import logging
+import secrets
+import sys
+import tempfile
+import textwrap
 import typing as t
 import unittest
 from unittest import mock
+from pathlib import Path
 
 from saq.job import CronJob, Job, Status
 from saq.utils import uuid1
-from saq.worker import Worker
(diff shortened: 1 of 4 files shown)


tests/test_worker.py
```diff
@@ -319,3 +325,64 @@ async def before_enqueue(job: Job) -> None:
         correlation_ids = await self.queue.apply("recurse", n=2)
         self.assertEqual(len(correlation_ids), 3)
         self.assertTrue(all(cid == correlation_ids[0] for cid in correlation_ids[1:]))
+
+
+class TestSettingsImport(unittest.TestCase):
```
tobymao, line 330: can you make this a new file?

pierec: I sure can!

tobymao: can you delete this one please?

Issue #119, opened 2024-05-25
"RuntimeError: read() called while another coroutine is already waiting for incoming data" while asyncio.gather job.refresh


evgenii-moriakhin:
I'm trying to setup saq for my work needs, and I've encountered this problem:

If try to execute asyncio.gather for job.refresh() with until_complete set, an error is raised

```
Failed to consume message
Traceback (most recent call last):
  File "/home/evgeniy/rq/saq/saq/queue.py", line 736, in _daemon
    message = await self.pubsub.get_message(timeout=None)  # type: ignore
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/client.py", line 927, in get_message
    response = await self.parse_response(block=(timeout is None), timeout=timeout)
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/client.py", line 804, in parse_response
    response = await self._execute(
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/client.py", line 784, in _execute
    return await conn.retry.call_with_retry(
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/retry.py", line 59, in call_with_retry
    return await do()
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/connection.py", line 782, in read_response
    response = await self._parser.read_response(
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/connection.py", line 411, in read_response
    await self.read_from_socket()
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/connection.py", line 392, in read_from_socket
    buffer = await self._stream.read(self._read_size)
  File "/usr/lib/python3.10/asyncio/streams.py", line 669, in read
    await self._wait_for_data('read')
  File "/usr/lib/python3.10/asyncio/streams.py", line 487, in _wait_for_data
    raise RuntimeError(
RuntimeError: read() called while another coroutine is already waiting for incoming data
```

here's the code that reproduces the problem

```python
import asyncio
from redis.asyncio import Redis
from saq.job import Job
from saq.queue import Queue
from saq.worker import Worker

def test(ctx):
    print("test")
    return 1

async def main():
    queue = Queue(Redis())
    worker = Worker(queue, [test])

    asyncio.create_task(worker.start())
    jobs = [Job("test") for _ in range(10)]
[...]

PR #121, opened 2024-05-25
add lock when creating pubsub daemon (#119)

evgenii-moriakhin:
#119

### DIFF
--- a/saq/queue.py
+++ b/saq/queue.py
@@ -729,6 +729,7 @@ class PubSubMultiplexer:
         self._subscriptions: t.Dict[bytes, t.Set[Q]] = defaultdict(set)
         self._queues: t.Dict[Q, t.Set[bytes]] = {}
         self._daemon_task: t.Optional[asyncio.Task] = None
+        self._daemon_creation_lock = asyncio.Lock()
 
     async def _daemon(self) -> None:
         while True:
@@ -743,9 +744,10 @@ class PubSubMultiplexer:
                 logger.exception("Failed to consume message")
 
     async def start(self) -> None:
-        if not self._daemon_task:
-            await self.pubsub.psubscribe(f"{self.prefix}*")
-            self._daemon_task = asyncio.create_task(self._daemon())
+        async with self._daemon_creation_lock:
+            if not self._daemon_task:
+                await self.pubsub.psubscribe(f"{self.prefix}*")
+                self._daemon_task = asyncio.create_task(self._daemon())
 
     async def close(self) -> None:
         await self.pubsub.punsubscribe()
--- a/saq/queue.py
+++ b/saq/queue.py
@@ -744,10 +744,11 @@ class PubSubMultiplexer:
                 logger.exception("Failed to consume message")
 
     async def start(self) -> None:
-        async with self._lock:
-            if not self._daemon_task:
-                await self.pubsub.psubscribe(f"{self.prefix}*")
-                self._daemon_task = asyncio.create_task(self._daemon())
+        if not self._daemon_task:
+            async with self._lock:
+                if not self._daemon_task:
+                    await self.pubsub.psubscribe(f"{self.prefix}*")
+                    self._daemon_task = asyncio.create_task(self._daemon())
 
     async def close(self) -> None:
         await self.pubsub.punsubscribe()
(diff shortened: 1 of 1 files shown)


saq/queue.py
```diff
@@ -729,6 +729,7 @@ def __init__(self, pubsub: PubSub, prefix: str):
         self._subscriptions: t.Dict[bytes, t.Set[Q]] = defaultdict(set)
         self._queues: t.Dict[Q, t.Set[bytes]] = {}
         self._daemon_task: t.Optional[asyncio.Task] = None
+        self._daemon_creation_lock = asyncio.Lock()
```
barakalon, line 732: Stylistically, we aim for concise names.

How about `self._creation_lock`, or even just `self._lock`, since this is the only one?

saq/queue.py
```diff
@@ -743,9 +744,10 @@ async def _daemon(self) -> None:
                 logger.exception("Failed to consume message")
 
     async def start(self) -> None:
-        if not self._daemon_task:
-            await self.pubsub.psubscribe(f"{self.prefix}*")
-            self._daemon_task = asyncio.create_task(self._daemon())
+        async with self._daemon_creation_lock:
```
barakalon, line 747: nit:

How about this?:
```python
if not self._daemon_task:
    async with self._lock:
        if not self._daemon_task:
            ...
```

Reasoning: its much quicker to check a boolean than it is to acquire lock. We expect this lock to only be needed once, so we're sacrificing some elegance for speed.

barakalon:
Great issue and PR - thanks for the contribution @evgenii-moriakhin !

saq/queue.py
evgenii-moriakhin: I've been thinking about it

Leads to the following problem

```
test
test
test
test
test
test
test
test
test
test
Failed to consume message
  File "/home/evgeniy/rq/saq/saq/queue.py", line 737, in _daemon
    message = await self.pubsub.get_message(timeout=None)  # type: ignore
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/client.py", line 1042, in get_message
    response = await self.parse_response(block=(timeout is None), timeout=timeout)
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/client.py", line 915, in parse_response
    response = await self._execute(
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/client.py", line 895, in _execute
    return await conn.retry.call_with_retry(
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/retry.py", line 62, in call_with_retry
    await fail(error)
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/client.py", line 884, in _disconnect_raise_connect
    raise error
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/retry.py", line 59, in call_with_retry
    return await do()
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/asyncio/connection.py", line 541, in read_response
    response = await self._parser.read_response(
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/_parsers/hiredis.py", line 206, in read_response
    await self.read_from_socket()
  File "/home/evgeniy/rq/venv/lib/python3.10/site-packages/redis/_parsers/hiredis.py", line 186, in read_from_socket
    raise ConnectionError(SERVER_CLOSED_CONNECTION_ERROR) from None
redis.exceptions.ConnectionError: Connection closed by server.
```
I assumed that it happens because I need to add same lock to close() as well.

But that doesn't solve the problem

```
    async def close(self) -> None:
        await self.pubsub.punsubscribe()
        if self._daemon_task:
            async with self._daemon_lifespan_lock:
                self._daemon_task.cancel()
                await self._daemon_task
                self._daemon_task = None
                self._subscriptions.clear()
                self._queues.clear()
```


any guesses on that?

saq/queue.py
evgenii-moriakhin: fix

saq/queue.py
barakalon: Stepping AFK for a bit, will take a look in a few hours

evgenii-moriakhin: Oops, I missed that you had the if not self._daemon_task: check twice. I only added it once, not realizing it doesn't fully solve the issue. With two checks, it looks good now.

barakalon (approved):

PR #130, opened 2024-08-01
Refactor Queue into an abstract base class…

vchan:
…and introduce a concrete Redis implementation.

### DIFF
--- a/saq/queue/base.py
+++ b/saq/queue/base.py
@@ -31,7 +31,6 @@ if t.TYPE_CHECKING:
         LoadType,
         QueueInfo,
         QueueStats,
-        VersionTuple,
     )
 
 
@@ -53,7 +52,6 @@ class JobError(Exception):
 class Queue(ABC):
     """An abstract base class for queues"""
 
-    @abstractmethod
     def __init__(
         self,
         name: str,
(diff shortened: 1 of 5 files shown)


saq/queue/base.py
```diff
@@ -0,0 +1,342 @@
...
+        super().__init__(
+            f"Job {job.id} {job.status}\n\nThe above job failed with the following error:\n\n{job.error}"
+        )
+        self.job = job
+
+
+class Queue(ABC):
+    """An abstract base class for queues"""
+
+    @abstractmethod
```
barakalon, line 56: I don’t think this is necessary

saq/queue/base.py
```diff
@@ -0,0 +1,342 @@
...
+        dump: DumpType | None,
+        load: LoadType | None,
+    ) -> None:
+        self.name = name
+        self.started: int = now()
+        self.complete = 0
+        self.failed = 0
+        self.retried = 0
+        self.aborted = 0
+        self._version: VersionTuple | None = None
```
barakalon, line 69: Is this originally the redis version? Does it belong here?

barakalon (approved):

saq/queue/base.py
vchan: Yes, this should be the redis version. I'll move it to the redis queue.

PR #131, opened 2024-08-08
Introduce Postgres queue

vchan:
This PR introduces a Postgres-backed queue. Still need to add tests and implement several missing features like info, stats, schedule, sweep, etc.

### DIFF
--- a/saq/queue/base.py
+++ b/saq/queue/base.py
@@ -95,6 +95,10 @@ class Queue(ABC):
         pass
 
     @abstractmethod
+    async def sweep(self, lock: int = 60, abort: float = 5.0) -> list[t.Any]:
+        pass
+
+    @abstractmethod
     async def listen(
         self,
         job_keys: Iterable[str],
@@ -150,18 +154,6 @@ class Queue(ABC):
     async def finish_abort(self, job: Job) -> None:
         pass
 
-    @abstractmethod
-    async def get_jobs_by_ids(self, job_ids: Iterable[str]) -> list[Job | None]:
-        pass
-
-    @abstractmethod
-    async def _get_active_jobs_for_sweep(self, lock: int) -> list[str]:
-        pass
-
-    @abstractmethod
-    async def _sweep_job(self, job_id: str) -> None:
-        pass
-
     @staticmethod
     def from_url(url: str, **kwargs: t.Any) -> Queue:
         """Create a queue with a redis url a name."""
@@ -185,49 +177,6 @@ class Queue(ABC):
 
     def unregister_before_enqueue(self, callback: BeforeEnqueueType) -> None:
         self._before_enqueues.pop(id(callback), None)
-
-    async def sweep(self, lock: int = 60, abort: float = 5.0) -> list[t.Any]:
-        job_ids = await self._get_active_jobs_for_sweep(lock)
-
-        swept = []
-        if job_ids:
-            for job_id, job in zip(job_ids, await self.get_jobs_by_ids(job_ids)):
-                if job:
-                    # in rare cases a sweep may occur between a dequeue and actual processing.
-                    # in that case, the job status is still set to queued, but only because it hasn't
-                    # had its status updated yet. try refreshing after a short sleep to pick up the latest status.
-                    if job.status == Status.QUEUED:
-                        await asyncio.sleep(0.1)
-                        await job.refresh()
-                        stuck = job.status == Status.QUEUED
-                    else:
-                        stuck = job.status != Status.ACTIVE or job.stuck
-
-                    if stuck:
-                        swept.append(job_id)
-                        await self.abort(job, error="swept")
-
-                        logger.info(
-                            "Sweeping job %s",
-                            job.info(logger.isEnabledFor(logging.DEBUG)),
-                        )
-
-                        if job.retryable:
-                            try:
-                                await job.refresh(abort)
-                            except asyncio.TimeoutError:
-                                logger.info("Could not abort job %s", job_id)
-                            finally:
-                                await self.retry(job, error="swept")
-                        else:
-                            await job.finish(Status.ABORTED, error="swept")
-                else:
-                    swept.append(job_id)
-
-                    await self._sweep_job(job_id)
-                    logger.info("Sweeping missing job %s", job_id)
-
-        return swept
 
     async def apply(
         self, job_or_func: str, timeout: float | None = None, **kwargs: t.Any
--- a/saq/queue/base.py
+++ b/saq/queue/base.py
@@ -379,3 +379,11 @@ class Queue(ABC):
 
         if status == Status.COMPLETE:
             job.progress = 1.0
+
+    def _update_stats(self, status: Status) -> None:
+        if status == Status.COMPLETE:
+            self.complete += 1
+        elif status == Status.FAILED:
+            self.failed += 1
+        elif status == Status.ABORTED:
+            self.aborted += 1
--- a/saq/queue/base.py
+++ b/saq/queue/base.py
@@ -380,6 +380,15 @@ class Queue(ABC):
         if status == Status.COMPLETE:
             job.progress = 1.0
 
+    def _get_stats(self) -> QueueStats:
+        return {
+            "complete": self.complete,
+            "failed": self.failed,
+            "retried": self.retried,
+            "aborted": self.aborted,
+            "uptime": now() - self.started,
+        }
+
     def _update_stats(self, status: Status) -> None:
         if status == Status.COMPLETE:
             self.complete += 1
--- a/saq/queue/base.py
+++ b/saq/queue/base.py
@@ -84,10 +84,6 @@ class Queue(ABC):
         pass
 
     @abstractmethod
-    async def stats(self, ttl: int = 60) -> QueueStats:
-        pass
-
-    @abstractmethod
     async def count(self, kind: CountKind) -> int:
         pass
 
@@ -125,11 +121,27 @@ class Queue(ABC):
         pass
 
     @abstractmethod
-    async def retry(self, job: Job, error: str | None) -> None:
+    async def dequeue(self, timeout: float = 0) -> Job | None:
         pass
 
     @abstractmethod
-    async def finish(
+    async def get_abort_errors(self, jobs: Iterable[Job]) -> list[bytes | None]:
+        pass
+
+    @abstractmethod
+    async def finish_abort(self, job: Job) -> None:
+        pass
+
+    @abstractmethod
+    async def write_stats(self, stats: QueueStats, ttl: int) -> None:
+        pass
+
+    @abstractmethod
+    async def _retry(self, job: Job, error: str | None) -> None:
+        pass
+
+    @abstractmethod
+    async def _finish(
         self,
         job: Job,
         status: Status,
@@ -140,19 +152,7 @@ class Queue(ABC):
         pass
 
     @abstractmethod
-    async def dequeue(self, timeout: float = 0) -> Job | None:
-        pass
-
-    @abstractmethod
-    async def enqueue(self, job_or_func: str | Job, **kwargs: t.Any) -> Job | None:
-        pass
-
-    @abstractmethod
-    async def get_abort_errors(self, jobs: Iterable[Job]) -> list[bytes | None]:
-        pass
-
-    @abstractmethod
-    async def finish_abort(self, job: Job) -> None:
+    async def _enqueue(self, job: Job) -> Job | None:
         pass
 
     @staticmethod
@@ -173,11 +173,103 @@ class Queue(ABC):
     async def connect(self) -> None:
         pass
 
+    async def stats(self, ttl: int = 60) -> QueueStats:
+        stats: QueueStats = {
+            "complete": self.complete,
+            "failed": self.failed,
+            "retried": self.retried,
+            "aborted": self.aborted,
+            "uptime": now() - self.started,
+        }
+
+        await self.write_stats(stats, ttl)
+        return stats
+
     def register_before_enqueue(self, callback: BeforeEnqueueType) -> None:
         self._before_enqueues[id(callback)] = callback
 
     def unregister_before_enqueue(self, callback: BeforeEnqueueType) -> None:
         self._before_enqueues.pop(id(callback), None)
+
+    async def retry(self, job: Job, error: str | None) -> None:
+        job.status = Status.QUEUED
+        job.error = error
+        job.completed = 0
+        job.started = 0
+        job.progress = 0
+        job.touched = now()
+
+        await self._retry(job=job, error=error)
+
+    async def finish(
+        self,
+        job: Job,
+        status: Status,
+        *,
+        result: t.Any = None,
+        error: str | None = None,
+    ) -> None:
+        job.status = status
+        job.result = result
+        job.error = error
+        job.completed = now()
+
+        if status == Status.COMPLETE:
+            job.progress = 1.0
+
+        await self._finish(job=job, status=status, result=result, error=error)
+
+        if status == Status.COMPLETE:
+            self.complete += 1
+        elif status == Status.FAILED:
+            self.failed += 1
+        elif status == Status.ABORTED:
+            self.aborted += 1
+
+    async def enqueue(self, job_or_func: str | Job, **kwargs: t.Any) -> Job | None:
+        """
+        Enqueue a job by instance or string.
+
+        Example:
+            .. code-block::
+
+                job = await queue.enqueue("add", a=1, b=2)
+                print(job.id)
+
+        Args:
+            job_or_func: The job or function to enqueue.
+                If a job instance is passed in, it's properties are overriden.
+            kwargs: Kwargs can be arguments of the function or properties of the job.
+
+        Returns:
+            If the job has already been enqueued, this returns None, else Job
+        """
+        job_kwargs: dict[str, t.Any] = {}
+
+        for k, v in kwargs.items():
+            if k in Job.__dataclass_fields__:  # pylint: disable=no-member
+                job_kwargs[k] = v
+            else:
+                job_kwargs.setdefault("kwargs", {})[k] = v
+
+        if isinstance(job_or_func, str):
+            job = Job(function=job_or_func, **job_kwargs)
+        else:
+            job = job_or_func
+
+            for k, v in job_kwargs.items():
+                setattr(job, k, v)
+
+        if job.queue and job.queue.name != self.name:
+            raise ValueError(f"Job {job} registered to a different queue")
+
+        job.queue = self
+        job.queued = now()
+        job.status = Status.QUEUED
+
+        await self._before_enqueue(job)
+
+        return await self._enqueue(job)
 
     async def apply(
         self, job_or_func: str, timeout: float | None = None, **kwargs: t.Any
@@ -333,66 +425,3 @@ class Queue(ABC):
     async def _before_enqueue(self, job: Job) -> None:
         for cb in self._before_enqueues.values():
             await cb(job)
-
-    def _create_job_for_enqueue(self, job_or_func: str | Job, **kwargs: t.Any) -> Job:
-        job_kwargs: dict[str, t.Any] = {}
-
-        for k, v in kwargs.items():
-            if k in Job.__dataclass_fields__:  # pylint: disable=no-member
-                job_kwargs[k] = v
-            else:
-                job_kwargs.setdefault("kwargs", {})[k] = v
-
-        if isinstance(job_or_func, str):
-            job = Job(function=job_or_func, **job_kwargs)
-        else:
-            job = job_or_func
-
-        if job.queue and job.queue.name != self.name:
-            raise ValueError(f"Job {job} registered to a different queue")
-
-        job.queue = self
-        job.queued = now()
-        job.status = Status.QUEUED
-
-        return job
-
-    def _update_job_for_retry(self, job: Job, error: str | None) -> None:
-        job.status = Status.QUEUED
-        job.error = error
-        job.completed = 0
-        job.started = 0
-        job.progress = 0
-        job.touched = now()
-
-    def _update_job_for_finish(
-        self,
-        job: Job,
-        status: Status,
-        result: t.Any = None,
-        error: str | None = None,
-    ) -> None:
-        job.status = status
-        job.result = result
-        job.error = error
-        job.completed = now()
-
-        if status == Status.COMPLETE:
-            job.progress = 1.0
-
-    def _get_stats(self) -> QueueStats:
-        return {
-            "complete": self.complete,
-            "failed": self.failed,
-            "retried": self.retried,
-            "aborted": self.aborted,
-            "uptime": now() - self.started,
-        }
-
-    def _update_stats(self, status: Status) -> None:
-        if status == Status.COMPLETE:
-            self.complete += 1
-        elif status == Status.FAILED:
-            self.failed += 1
-        elif status == Status.ABORTED:
-            self.aborted += 1
--- a/saq/queue/base.py
+++ b/saq/queue/base.py
@@ -173,6 +173,18 @@ class Queue(ABC):
     async def connect(self) -> None:
         pass
 
+    def serialize(self, job: Job) -> bytes | str:
+        return self._dump(job.to_dict())
+
+    def deserialize(self, job_bytes: bytes | None) -> Job | None:
+        if not job_bytes:
+            return None
+
+        job_dict = self._load(job_bytes)
+        if job_dict.pop("queue") != self.name:
+            raise ValueError(f"Job {job_dict} fetched by wrong queue: {self.name}")
+        return Job(**job_dict, queue=self)
+
     async def stats(self, ttl: int = 60) -> QueueStats:
         stats: QueueStats = {
             "complete": self.complete,
--- a/saq/queue/base.py
+++ b/saq/queue/base.py
@@ -95,10 +95,6 @@ class Queue(ABC):
         pass
 
     @abstractmethod
-    async def sweep(self, lock: int = 60, abort: float = 5.0) -> list[t.Any]:
-        pass
-
-    @abstractmethod
     async def listen(
         self,
         job_keys: Iterable[str],
@@ -147,11 +143,23 @@ class Queue(ABC):
         pass
 
     @abstractmethod
-    async def get_many(self, keys: Iterable[str]) -> list[bytes | None]:
+    async def get_abort_errors(self, jobs: Iterable[Job]) -> list[bytes | None]:
         pass
 
     @abstractmethod
-    async def delete(self, key: str) -> None:
+    async def finish_abort(self, job: Job) -> None:
+        pass
+
+    @abstractmethod
+    async def get_jobs_by_ids(self, job_ids: Iterable[str]) -> list[Job | None]:
+        pass
+
+    @abstractmethod
+    async def _get_active_jobs_for_sweep(self, lock: int) -> list[str]:
+        pass
+
+    @abstractmethod
+    async def _sweep_job(self, job_id: str) -> None:
         pass
 
     @staticmethod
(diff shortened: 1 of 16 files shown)


saq/queue/postgres.py
```diff
@@ -0,0 +1,334 @@
...
+        LoadType,
+        QueueInfo,
+        QueueStats,
+    )
+
+JOBS_TABLE = "saq_jobs"
+
+CREATE_JOBS_TABLE = """
+CREATE TABLE IF NOT EXISTS saq_jobs (
+    key TEXT PRIMARY KEY,
```
tobymao, line 38: we'll need some other indices, like something that partitions on status, scheduled, etc

saq/queue/postgres.py
```diff
@@ -0,0 +1,334 @@
...
+        max_size: int = 5,
+    ) -> None:
+        super().__init__(name=name, dump=dump, load=load)
+
+        self.pool = pool
+        self.max_size = max_size
+
+    async def connect(self) -> None:
+        await self.pool.open()
+        await self.pool.resize(min_size=0, max_size=self.max_size)
```
tobymao, line 83: what does resize min_size=0 do?

saq/queue/postgres.py
```diff
@@ -0,0 +1,334 @@
...
+        if kind == "active":
+            status = [Status.ACTIVE]
+        if kind == "incomplete":
+            status = [Status.NEW, Status.DEFERRED, Status.QUEUED, Status.ACTIVE]
+
+        if status:
+            async with self.pool.connection() as conn, conn.cursor() as cursor:
+                await cursor.execute(
+                    f"""
+                    SELECT count(*) FROM {JOBS_TABLE}
```
tobymao, line 127: should we use %() for everything here?

saq/queue/postgres.py
```diff
@@ -0,0 +1,334 @@
...
+        return []
+
+    async def listen(
+        self,
+        job_keys: Iterable[str],
+        callback: ListenCallback,
+        timeout: float | None = 10,
+    ) -> None:
+        async with self.pool.connection() as conn:
+            await conn.set_autocommit(True)
```
tobymao, line 149: why do you manually neet to set this?

saq/queue/postgres.py
```diff
@@ -0,0 +1,334 @@
...
+        logger.info("Finished %s", job.info(logger.isEnabledFor(logging.DEBUG)))
+
+    async def dequeue(self, timeout: float = 0) -> Job | None:
+        async with self.pool.connection() as conn, conn.cursor() as cursor:
+            await cursor.execute(
+                f"""
+                UPDATE {JOBS_TABLE} SET status = %(active)s
+                WHERE key = (
+                  SELECT key FROM {JOBS_TABLE}
+                  WHERE status = %(queued)s AND queue = %(queue)s and EXTRACT(EPOCH FROM now()) >= scheduled
```
tobymao, line 270: these are the indices to care about

saq/queue/postgres.py
```diff
@@ -0,0 +1,334 @@
...
+                DELETE FROM {JOBS_TABLE}
+                WHERE key = %(key)s
+                """,
+                {"key": key},
+            )
+
+    def _escape_channel_name(self, key: str) -> str:
+        return f"channel_{key.replace('-', '_')}"
+
+    def _unescape_channel_name(self, channel: str) -> str:
```
tobymao, line 333: why do you need this?

saq/queue/redis.py
```diff
@@ -340,13 +340,8 @@ async def abort(self, job: Job, error: str, ttl: float = 5) -> None:
                 await self.redis.lrem(self._active, 0, job.id)
 
     async def retry(self, job: Job, error: str | None) -> None:
+        await super().retry(job, error)
```
tobymao, line 343: can we make it so you never have to call super? and then the subclasses implement _retry() or something

setup.py
```diff
@@ -36,11 +36,13 @@
     ],
     extras_require={
         "hiredis": ["redis[hiredis]>=4.2.0"],
+        "postgres": ["psycopg[pool]"],
```
tobymao, line 39: let's remove redis from install_requires line 34

saq/queue/postgres.py
```diff
@@ -0,0 +1,334 @@
...
+        ListenCallback,
+        DumpType,
+        LoadType,
+        QueueInfo,
+        QueueStats,
+    )
+
+JOBS_TABLE = "saq_jobs"
+
+CREATE_JOBS_TABLE = """
```
barakalon, line 36: 1. Some applications might not have permission to create tables, so this should probably be made optional and executable by a standalone process
2. Should we be thinking ahead about migrations? What if we want to add a column, etc. Perhaps we should be syncing a sequence of changes.

barakalon: Backwards compatible migrations, that don’t require users to go in and alter their DB, that is

tobymao: i think it's ok that we always create the table, but i'd make it a tad more configurable like what the table name is?

barakalon: So if we ever need to change we expect users to update the table name?

saq/queue/postgres.py
vchan: Postgres channel names have the same restrictions as table names. This was a hack to work around it. I'll add a better way to encode channel names.

saq/queue/postgres.py
vchan: For AsyncNullConnectionPool, min_size is always 0 since it doesn't maintain any connections. I've switched the queue to use an AsyncConnectionPool.

saq/queue/postgres.py
vchan: Realized I could just quote the channel name and not worry about this...

saq/queue/postgres.py
vchan: Due to the way notifications are implemented in Postgres, if I don't set this, the LISTEN will be in a transaction and won't receive any notifications until the transaction ends.

saq/queue/base.py
```diff
@@ -147,37 +143,91 @@ async def enqueue(self, job_or_func: str | Job, **kwargs: t.Any) -> Job | None:
...
-
-        job_dict = self._load(job_bytes)
-        if job_dict.pop("queue") != self.name:
-            raise ValueError(f"Job {job_dict} fetched by wrong queue: {self.name}")
-        return Job(**job_dict, queue=self)
+    async def sweep(self, lock: int = 60, abort: float = 5.0) -> list[t.Any]:
+        job_ids = await self._get_active_jobs_for_sweep(lock)
+
+        swept = []
+        if job_ids:
```
tobymao, line 193: this is redis specific hack, i don't think we need this for postgres right

saq/queue/postgres.py
```diff
@@ -0,0 +1,447 @@
...
+    def deserialize(self, job_dict: dict[t.Any, t.Any]) -> Job | None:
+        if job_dict.pop("queue") != self.name:
+            raise ValueError(f"Job {job_dict} fetched by wrong queue: {self.name}")
+        return Job(**job_dict, queue=self)
+
+    async def disconnect(self) -> None:
+        await self.pool.close()
+
+    async def info(
+        self, jobs: bool = False, offset: int = 0, limit: int = 10
```
tobymao, line 117: doesn't look like jobs, offset, or limit is used

saq/queue/postgres.py
tobymao: shouldn't this be self.jobs_table?

saq/queue/postgres.py
```diff
@@ -0,0 +1,447 @@
...
+                    SELECT count(*) FROM {JOBS_TABLE}
+                    WHERE status = 'active'
+                      AND queue = %(queue)s
+                    """,
+                    {"queue": self.name},
+                )
+            elif kind == "incomplete":
+                await cursor.execute(
+                    f"""
+                    SELECT count(*) FROM {JOBS_TABLE}
```
tobymao, line 161: ditto

saq/queue/postgres.py
```diff
@@ -0,0 +1,447 @@
...
+            if job:
+                return self.deserialize(job[0])
+        return None
+
+    async def abort(self, job: Job, error: str, ttl: float = 5) -> None:
+        job.error = error
+        job.status = Status.ABORTING
+
+        await self.update(job)
+        asyncio.create_task(self.delete(job, status=Status.ABORTED, delay=ttl))
```
tobymao, line 239: why create_task here?

where is the code that actually kills the task?

saq/queue/postgres.py
```diff
@@ -0,0 +1,447 @@
...
+                    f"""
+                    UPDATE {self.jobs_table} SET status = 'active'
+                    WHERE key = (
+                      SELECT key
+                      FROM {self.jobs_table}
+                      WHERE status = 'queued'
+                        AND queue = %(queue)s
+                        AND %(now)s >= (queued + scheduled)
+                      FOR UPDATE SKIP LOCKED
+                      LIMIT 1
```
tobymao, line 326: shouldn't limit 1 be determined by the concurrency argument?

saq/queue/postgres.py
```diff
@@ -0,0 +1,447 @@
...
+            if not await cursor.fetchone():
+                return None
+
+        logger.info("Enqueuing %s", job.info(logger.isEnabledFor(logging.DEBUG)))
+        return job
+
+    async def delete(
+        self, job: Job, status: Status | None = None, delay: float = 0
+    ) -> None:
+        if delay:
```
tobymao, line 383: why do we need a delay for delete?

saq/queue/postgres.py
vchan: That's a typo.

saq/queue/postgres_ddl.py
```diff
@@ -0,0 +1,15 @@
...
+    queued BIGINT NOT NULL DEFAULT EXTRACT(EPOCH FROM now()),
+    job JSONB,
+    queue TEXT,
+    status TEXT,
+    scheduled BIGINT
+);
+"""
+
+CREATE_JOBS_DEQUEUE_INDEX = """
+CREATE INDEX IF NOT EXISTS saq_jobs_dequeue_idx ON {jobs_table} (status, queue, (queued + scheduled))
```
tobymao, line 14: does this index work? queued + scheduled? does that get hit?

saq/queue/postgres.py
vchan: This deletes aborted jobs from the jobs table. The worker contains the code that kills the task.

saq/queue/postgres_ddl.py
tobymao: why do we have to do math? is that how it was done? is scheduled not just a timestamp it's an integer seconds?

saq/queue/postgres.py
vchan: I think we need a janitor task to clear out jobs from the jobs table based on ttl, so this can go away.

saq/queue/postgres.py
vchan: This is tied to the ttl for removing completed (including aborted) tasks. I'll add a janitor task to replace this though.

saq/queue/postgres.py
```diff
@@ -0,0 +1,521 @@
...
+
+
+class PostgresQueue(Queue):
+    """
+    Queue is used to interact with Postgres.
+
+    Args:
+        pool: instance of psycopg_pool.AsyncConnectionPool
+        name: name of the queue (default "default")
+        joabs_table: name of the Postgres table SAQ will write jobs to (default "saq_jobs")
```
tobymao, line 54: jobs_table

saq/queue/postgres.py
```diff
@@ -0,0 +1,521 @@
...
+            json.set_json_loads(load)
+
+        self.connections: dict[str, AsyncConnection] = {}
+        self.cond = asyncio.Condition()
+
+    async def connect(self) -> None:
+        await self.pool.open()
+        await self.pool.resize(min_size=self.min_size, max_size=self.max_size)
+        async with self.pool.connection() as conn, conn.cursor() as cursor:
+            await cursor.execute(CREATE_JOBS_TABLE.format(jobs_table=self.jobs_table))
```
tobymao, line 103: can you do a for loop on all of the init scripts?

saq/queue/postgres.py
```diff
@@ -0,0 +1,521 @@
...
+                f"""
+                SELECT job
+                FROM {self.jobs_table} LEFT OUTER JOIN pg_locks ON lock_id = objid
+                WHERE status = 'active'
+                  AND objid IS NULL
+                """
+            )
+            results = await cursor.fetchall()
+        swept = []
+        for result in results:
```
tobymao, line 218: what happens if the sweeper gets killed after the delete from jobs table but before retry / abort?

saq/queue/postgres.py
```diff
@@ -0,0 +1,521 @@
...
+    async def _notify(self, channel: str, payload: t.Any) -> None:
+        async with self.pool.connection() as conn:
+            await conn.execute(f"NOTIFY \"{channel}\", '{payload}'")
+
+    async def update(self, job: Job) -> None:
+        job.touched = now()
+        async with self._get_connection(job.key) as conn:
+            await conn.execute(
+                f"""
+                    UPDATE {self.jobs_table} SET job = %(job)s, status = %(status)s
```
tobymao, line 268: are you unable to pass in self.jobs_table as a param? what if thehre's sql injection in there

saq/queue/postgres.py
```diff
@@ -0,0 +1,521 @@
...
+            )
+            job = await cursor.fetchone()
+            if job:
+                return self.deserialize(job[0])
+        return None
+
+    async def abort(self, job: Job, error: str, ttl: float = 5) -> None:
+        job.error = error
+        job.status = Status.ABORTING
+        job.ttl = int(seconds(now()) + ttl)
```
tobymao, line 292: why are you setting the job ttl here? i don't think redis does this. redis usses the abort ttl for the abort key right?

saq/queue/postgres.py
```diff
@@ -0,0 +1,521 @@
...
+        self._update_job_for_retry(job, error)
+        next_retry_delay = job.next_retry_delay()
+        if next_retry_delay:
+            scheduled = time.time() + next_retry_delay
+        else:
+            scheduled = job.scheduled or seconds(now())
+
+        async with self._get_connection(job.key) as conn:
+            await conn.execute(
+                f"""
```
tobymao, line 306: should this just leverage the update code? and update should better handle passing in scheduled?

tobymao:
```
def update(self, job, status=..., scheduled=...)
```

saq/queue/postgres.py
```diff
@@ -0,0 +1,521 @@
...
+        await job.finish(Status.ABORTED, error=job.error)
+
+    async def dequeue_timer(self, poll_interval: int) -> None:
+        """Wakes up a single dequeue task every `poll_interval` seconds."""
+        while True:
+            async with self.cond:
+                self.cond.notify(1)
+            await asyncio.sleep(poll_interval)
+
+    async def wait_for_job(self) -> Job:
```
tobymao, line 432: i feel like this isn't necessary, it's pretty much the same as dequeue, can we move it inline into dequeu?

saq/queue/postgres.py
```diff
@@ -0,0 +1,521 @@
...
+                await self.cond.wait()
+            job = await self._dequeue()
+            if job:
+                break
+        return job
+
+    async def listen_for_enqueues(self, timeout: float | None = None) -> None:
+        """Wakes up a single dequeue task when a Postgres enqueue notification is received."""
+        async with self.pool.connection() as conn:
+            await conn.execute('LISTEN "saq:enqueue"')
```
tobymao, line 449: saq:enqueue make constant?

saq/queue/postgres.py
vchan: Do you mean make this an inner function in dequeue()? This was added so dequeue() can use asyncio.wait_for() for timing out, but this can be nested inside dequeue().

tobymao: yea lambda inside

saq/queue/postgres.py
vchan: Redis also sets the job expiration to abort ttl + 1. I can update this to match what Redis is doing.

saq/queue/postgres.py
vchan: The delete above is for clearing out jobs with terminal statuses past their ttl, so it's not related to retry/abort. However, all of sweep now takes place in a transaction, so everything should be rolled back if the sweeper gets killed before it finishes.

saq/queue/postgres.py
```diff
@@ -0,0 +1,608 @@
...
+    async def sweep(self, lock: int = 60, abort: float = 5.0) -> list[t.Any]:
+        """Delete jobs past their TTL from the jobs table"""
+        swept = []
+
+        async with self.pool.connection() as conn, conn.cursor() as cursor:
+            await cursor.execute(
+                """
+            SELECT pg_try_advisory_xact_lock(%(classid)s, %(objid)s)
+            """,
+                {"classid": SWEEP_LOCK_KEY[0], "objid": SWEEP_LOCK_KEY[1]},
```
tobymao, line 271: why is the sweep lock key 1, 1, how did you decide that?

saq/queue/postgres.py
```diff
@@ -0,0 +1,608 @@
...
+                """
+                ).format(jobs_table=self.jobs_table),
+                {"now": math.ceil(seconds(now()))},
+            )
+            await cursor.execute(
+                SQL(
+                    """
+                SELECT job
+                FROM {jobs_table} LEFT OUTER JOIN pg_locks ON lock_key = objid
+                WHERE status = 'active'
```
tobymao, line 293: so this finds jobs that are active but not locked, so you know they are stuck right?

saq/queue/postgres.py
```diff
@@ -0,0 +1,608 @@
...
+        self, job: Job, status: Status | None = None, scheduled: int | None = None
+    ) -> None:
+        if status:
+            job.status = status
+        if scheduled:
+            job.scheduled = scheduled
+        job.touched = now()
+
+        async with self.pool.connection() as conn:
+            await conn.execute(
```
tobymao, line 355: is this done in a transaction? can we also notify in the transaction too so that you'll never update a job and skip a notification?

saq/queue/postgres.py
```diff
@@ -0,0 +1,608 @@
...
+                {"key": job_key},
+            )
+            job = await cursor.fetchone()
+            if job:
+                return self.deserialize(job[0])
+        return None
+
+    async def abort(self, job: Job, error: str, ttl: float = 5) -> None:
+        job.error = error
+        job.status = Status.ABORTING
```
tobymao, line 389: where is the job actually aborted?

saq/queue/postgres.py
```diff
@@ -0,0 +1,608 @@
...
+
+        async with self.pool.connection() as conn, conn.cursor() as cursor:
+            await cursor.execute(
+                SQL(
+                    """
+                INSERT INTO {jobs_table} (key, job, queue, status, scheduled)
+                VALUES (%(key)s, %(job)s, %(queue)s, %(status)s, %(scheduled)s)
+                ON CONFLICT (key) DO UPDATE
+                SET job = %(job)s, queue = %(queue)s, status = %(status)s, scheduled = %(scheduled)s
+                WHERE {jobs_table}.status NOT IN ('queued', 'aborting')
```
tobymao, line 490: so we won't update jobs that are queued or aborting, but we will update them if they are active???

saq/queue/postgres.py
```diff
@@ -0,0 +1,608 @@
...
+        return job
+
+    async def get_abort_errors(self, jobs: Iterable[Job]) -> list[bytes | None]:
+        async with self.pool.connection() as conn, conn.cursor() as cursor:
+            await cursor.execute(
+                SQL(
+                    """
+                SELECT key, job->'error'
+                FROM {jobs_table}
+                WHERE key = ANY(%(keys)s)
```
tobymao, line 518: is this always sorted automatically? does the order matter?

saq/queue/postgres.py
```diff
@@ -0,0 +1,608 @@
...
+                  ORDER BY scheduled
+                  LIMIT %(limit)s
+                  FOR UPDATE SKIP LOCKED
+                )
+                UPDATE {jobs_table} SET status = 'active'
+                FROM locked_job
+                WHERE {jobs_table}.key = locked_job.key AND pg_try_advisory_lock(locked_job.lock_key)
+                RETURNING job
+                """
+                ).format(jobs_table=self.jobs_table),
```
tobymao, line 573: why is jobs table passed in with format and not the params below? we shouldn't use old style over f-strings, that's not what i meant

saq/queue/postgres.py
```diff
@@ -0,0 +1,608 @@
...
+            return
+        async with self._get_connection() as conn:
+            await conn.execute(
+                SQL(
+                    """
+                SELECT pg_advisory_unlock(lock_key)
+                FROM {jobs_table}
+                WHERE key = ANY(%(keys)s)
+                """
+                ).format(jobs_table=self.jobs_table),
```
tobymao, line 604: same here

saq/queue/postgres.py
```diff
@@ -0,0 +1,608 @@
...
+                RETURNING key
+                """
+                ).format(jobs_table=self.jobs_table),
+                {"now": math.ceil(seconds(now()))},
+            )
+            await cursor.execute(
+                SQL(
+                    """
+                SELECT job
+                FROM {jobs_table} LEFT OUTER JOIN pg_locks ON lock_key = objid
```
tobymao, line 292: how does lock_key = objid if objid is null? null never == null

saq/queue/postgres.py
```diff
@@ -0,0 +1,608 @@
...
+    async def stats(self, ttl: int = 60) -> QueueStats:
+        stats = self._get_stats()
+        async with self.pool.connection() as conn:
+            await conn.execute(
+                SQL(
+                    """
+                INSERT INTO {stats_table} (worker_id, stats, ttl)
+                VALUES (%(worker_id)s, %(stats)s, %(ttl)s)
+                ON CONFLICT (worker_id) DO UPDATE
+                SET stats = %(stats)s, ttl = %(ttl)s
```
tobymao, line 206: is the stats table ever cleaned up? what does the ttl do?

vchan: TTL is used by info to filter out stale stats. I'll also add something in the sweeper to delete old stats.

saq/queue/postgres.py
vchan: I just picked it arbitrarily. I didn't want it to potentially collide with the job lock keys.

saq/queue/postgres.py
vchan: Right, but there could be null objids because it's a left outer join if some jobs don't have locks.

saq/queue/postgres.py
vchan: Yes, exactly.

saq/queue/postgres.py
tobymao: what does redis do?

saq/queue/postgres.py
vchan: The worker's abort function cancels the task and calls queue.finish_abort(), which updates the status to aborted.

saq/queue/postgres.py
tobymao: can you look through all the places and make sure things are as robust as possible with using transactions?

saq/queue/postgres.py
vchan: This is calling format on a psycopg.sql.SQL object, not on a string. This is the psycopg way to dynamically set tables and fields for a query and prevent SQL injection. We can't bind tables and fields in queries like other query values with placeholders like with the params below.

tobymao: great

saq/queue/postgres.py
vchan: Good point. I think we should only update, aka re-queue, jobs that are done, i.e. finished, failed, or aborted. What do you think?

tobymao: TERMINAL_STATUSES right?

saq/queue/postgres.py
vchan: The Redis queue:

1. Saves the current worker's stat to key "saq:stats:{worker_id}" with an expiration of ttl.
2. Removes stale worker_ids in the sorted set "saq:stats".
3. Add the current worker's id to the sorted set "saq:stats".
4. Sets an expiration of ttl on the sorted set "saq:stats".

Since we can't automatically expire rows in our stats tables, I'm using ttl to filter out stale stats in info() and add a step in the sweeper to delete them.

saq/queue/postgres.py
vchan: Yes, TERMINAL_STATUSES

saq/queue/postgres.py
vchan: Yes, order does matter. No, I don't think order is guaranteed here. There's additional logic below to reorder the errors to match the input jobs list.

saq/queue/postgres.py
```diff
@@ -0,0 +1,648 @@
...
+
+        self.cond = asyncio.Condition()
+        self.queue: asyncio.Queue = asyncio.Queue()
+        self.waiting = 0  # Internal counter of worker tasks waiting for dequeue
+        self._connected = False
+        self.connection_lock = asyncio.Lock()
+        self.released: list[str] = []
+
+    async def connect(self) -> None:
+        if self._connected:
```
tobymao, line 111: let's remove self._connected and look at self.connection or self.pool instead

saq/queue/postgres.py
tobymao: it doesn't seem like you're using the lock timer here, we only want to sweep every 60 seconds no matter how many workers there are right

saq/queue/postgres.py
tobymao: what about jobs that are stuck in 'aborting' ?

saq/queue/postgres.py
```diff
@@ -0,0 +1,648 @@
...
+        status: Status,
+        *,
+        result: t.Any = None,
+        error: str | None = None,
+    ) -> None:
+        self._update_job_for_finish(job, status, result=result, error=error)
+        key = job.key
+
+        async with self.pool.connection() as conn, conn.cursor() as cursor:
+            if job.ttl >= 0:
```
tobymao, line 457: hey, this doesn't match redis's behavior, if ttl > 0, then set expiration, if it's 0, there is no expiration, else delete. right?

saq/queue/redis.py
```diff
@@ -392,12 +392,7 @@ async def finish(
...
 
-            if status == Status.COMPLETE:
-                self.complete += 1
-            elif status == Status.FAILED:
-                self.failed += 1
-            elif status == Status.ABORTED:
-                self.aborted += 1
+            self._update_stats(status)
 
             await self.notify(job)
```
tobymao, line 397: why is conn passed in some places but not others?

vchan: This is the Redis queue, so there's no conn to pass here. In Postgres, I've centralized most calls to notify() in update() and the remaining call to notify() in finish() passes conn. We generally want conn so notify can be part of conn's transaction.

saq/queue/postgres.py
vchan: Good catch. This is wrong after I refactored finish() to use update(). I've fixed it.

saq/queue/postgres.py
```diff
@@ -439,7 +439,6 @@ async def retry(self, job: Job, error: str | None) -> None:
         await self._release_job(job.key)
 
         self.retried += 1
-        await self.notify(job)
```
tobymao: why is this removed?

vchan: The update() function already calls notify().

saq/queue/postgres.py
vchan: Ah, you're right. Need to update this.

saq/queue/postgres.py
```diff
@@ -0,0 +1,648 @@
...
+                    """
+                DELETE FROM {jobs_table}
+                WHERE status IN ('aborted', 'complete', 'failed')
+                  AND %(now)s >= ttl
+                """
+                ).format(jobs_table=self.jobs_table),
+                {"now": math.ceil(seconds(now()))},
+            )
+            # Delete expired stats
+            await cursor.execute(
```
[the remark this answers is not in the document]
vchan: Yes, I've update this to combine them now.

saq/queue/postgres.py
vchan: Yes, if the worker aborting the job dies before it can update the job's status. I'll change this to also sweep jobs stuck in 'aborting`.

tobymao: make sure we don’t sweep jobs that are actually aborting

vchan: Okay, I've double checked the sweeper makes sure there's no lock associated with a job before sweeping, and aborting releases the lock after canceling the task and updating the status to 'aborted'.

setup.py
```diff
@@ -31,17 +31,20 @@
...
+        "redis": ["redis>=4.2,<6.0"],
         "web": ["aiohttp", "aiohttp_basicauth"],
         "dev": [
             "black",
             "coverage",
             "mypy",
+            "psycopg[pool]",
```
barakalon, line 45: We should just update the `make deps` command to include redis and postgres: https://github.com/tobymao/saq/blob/master/Makefile#L2

tests/helpers.py
```diff
@@ -1,13 +1,32 @@
...
+    await queue.connect()
+    return queue
+
+
+async def cleanup_queue(queue: Queue) -> None:
+    if isinstance(queue, RedisQueue):
+        await queue.redis.flushdb()
+    elif isinstance(queue, PostgresQueue):
+        async with queue.pool.connection() as conn:
+            await conn.execute("DROP SCHEMA public CASCADE")
```
barakalon, line 28: Yikes. Should we be using a dedicated test schema?

saq/queue/postgres.py
```diff
@@ -0,0 +1,667 @@
...
+        self.has_sweep_lock = False
+
+    async def info(
+        self, jobs: bool = False, offset: int = 0, limit: int = 10
+    ) -> QueueInfo:
+        async with self.pool.connection() as conn, conn.cursor() as cursor:
+            await cursor.execute(
+                SQL(
+                    """
+                SELECT worker_id, stats FROM {stats_table}
```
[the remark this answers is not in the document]
vchan: I passed all of the multiline SQL strings to `textwrap.dedent`. Wanted to confirm that is what you meant.

PR #141, opened 2024-09-03
Fix cronjob inability to pass kwargs to the function

Seemone:
Fix a bug where cronjob creation could not accept a kwargs parameter as described in the documentation


barakalon:
Where in the documentation does it say CronJobs accept kwargs?

tobymao:
it says here 

```
    **The remaining kwargs are pass through to Job:** (see equivalent field in Job for more)

    * timeout
    * heartbeat
    * retries
    * ttl
```

not arbitrary kwargs

Seemone:
@tobymao tests amended

you're right, I misread the comment.

consider this a feature addition then

PR #145, opened 2024-09-06
Chore: Introduce queue upkeep

vchan:
This PR introduces queue specific upkeep tasks. For the Postgres queue, this helps decouple opening the queue's connection pool and various upkeep tasks like the dequeue timer and enqueue listener.

### DIFF
--- a/saq/queue/postgres.py
+++ b/saq/queue/postgres.py
@@ -120,14 +120,13 @@ class PostgresQueue(Queue):
                     )
                 )
 
-    async def upkeep(self) -> set[asyncio.Task[None]]:
+    async def upkeep(self) -> None:
         await self.init_db()
 
         self.tasks.add(asyncio.create_task(self.wait_for_job()))
         self.tasks.add(asyncio.create_task(self.listen_for_enqueues()))
         if self.poll_interval > 0:
             self.tasks.add(asyncio.create_task(self.dequeue_timer(self.poll_interval)))
-        return self.tasks
 
     async def connect(self) -> None:
         if self.connection:
--- a/saq/queue/postgres.py
+++ b/saq/queue/postgres.py
@@ -110,7 +110,24 @@ class PostgresQueue(Queue):
         self.connection_lock = asyncio.Lock()
         self.released: list[str] = []
         self.has_sweep_lock = False
-        self.tasks: list[asyncio.Task] = []
+
+    async def init_db(self) -> None:
+        async with self.pool.connection() as conn, conn.cursor() as cursor:
+            for statement in DDL_STATEMENTS:
+                await cursor.execute(
+                    SQL(statement).format(
+                        jobs_table=self.jobs_table, stats_table=self.stats_table
+                    )
+                )
+
+    async def upkeep(self) -> set[asyncio.Task[None]]:
+        await self.init_db()
+
+        self.tasks.add(asyncio.create_task(self.wait_for_job()))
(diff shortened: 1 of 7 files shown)


tests/test_queue.py
```diff
@@ -176,7 +176,7 @@ async def test_retry(self) -> None:
 
     async def test_retry_delay(self) -> None:
         # Let's first verify how things work without a retry delay
-        worker = Worker(self.queue, functions=functions, dequeue_timeout=0.01)
+        worker = Worker(self.queue, functions=functions, dequeue_timeout=0.1)
```
vchan, line 179: The low timeout was causing this test to be flaky since the Postgres queue isn't as fast as the Redis one.

tests/test_queue.py
```diff
@@ -440,12 +440,14 @@ async def asyncTearDown(self) -> None:
         await super().asyncTearDown()
         await teardown_postgres()
 
+    @unittest.skip("Not implemented")
```
vchan, line 443: Skipping with the decorator also skips setUp and tearDown.

tests/helpers.py
```diff
@@ -24,12 +25,18 @@ async def create_postgres_queue(**kwargs: t.Any) -> PostgresQueue:
         ),
     )
     await queue.connect()
+    await queue.upkeep()
+    await asyncio.sleep(0.1)  # Give some time for the tasks to start
```
vchan, line 29: Without this, some tests were flaky because tearDown would be called before some upkeep tasks could start.

saq/worker.py
```diff
@@ -341,6 +340,13 @@ def start(
     loop = asyncio.new_event_loop()
     worker = Worker(**settings_obj)
 
+    async def worker_start() -> None:
```
tobymao, line 343: why did you move this? isn't worker.stop already called on the finally of def start?

saq/queue/postgres.py
```diff
@@ -110,31 +110,34 @@ def __init__(
...
                 )
-        # Reserve a connection for dequeue and advisory locks
-        self.connection = await self.pool.getconn()
 
-        self.tasks.append(asyncio.create_task(self.wait_for_job()))
-        self.tasks.append(asyncio.create_task(self.listen_for_enqueues()))
+    async def upkeep(self) -> set[asyncio.Task[None]]:
+        await self.init_db()
+
+        self.tasks.add(asyncio.create_task(self.wait_for_job()))
```
tobymao, line 126: it doesn't look like queue.tasks is used anymore, can we remove it? or do we actually need to use it?

tests/helpers.py
```diff
@@ -24,12 +25,18 @@ async def create_postgres_queue(**kwargs: t.Any) -> PostgresQueue:
...
     return queue
 
 
 async def cleanup_queue(queue: Queue) -> None:
     if isinstance(queue, RedisQueue):
         await queue.redis.flushdb()
+    for task in queue.tasks:
```
tobymao, line 36: i see queue.tasks is only used here? that seems like a smell

saq/worker.py
vchan: I wanted to decouple queue connect/disconnect from worker start/stop. That way, if there are multiple workers sharing the same queue, stopping one won't disconnect the queue for the others. This happens in the tests, for example.

saq/queue/postgres.py
vchan: I gave Queue its own stop() method.

tobymao (approved):

PR #149, opened 2024-09-09
Fix!: Postgres queue sweep

vchan:
Fix the Postgres queue's sweep to match the Redis queue

### DIFF
--- a/saq/queue/postgres.py
+++ b/saq/queue/postgres.py
@@ -289,10 +289,9 @@ class PostgresQueue(Queue):
                     AND classid = {job_lock_keyspace}
                     AND objsubid = 2 -- key is int pair, not single bigint
                 )
-                SELECT job
+                SELECT key, job, objid
                 FROM {jobs_table} LEFT OUTER JOIN locks ON lock_key = objid
-                WHERE status IN ('active', 'aborting')
-                  AND objid IS NULL;
+                WHERE status IN ('active', 'aborting');
                 """
                     )
                 ).format(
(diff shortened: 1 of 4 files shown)


saq/queue/postgres.py
```diff
@@ -306,11 +305,42 @@ async def sweep(self, lock: int = 60, abort: float = 5.0) -> list[str]:
             cursor.nextset()
             cursor.nextset()
             results = await cursor.fetchall()
-            for result in results:
-                job = self.deserialize(result[0])
+            for key, job_bytes, objid in results:
+                job = self.deserialize(job_bytes)
                 if not job:
```
tobymao, line 310: how can job be null?

saq/queue/postgres.py
```diff
@@ -306,11 +305,42 @@ async def sweep(self, lock: int = 60, abort: float = 5.0) -> list[str]:
...
+                                DELETE FROM {jobs_table}
+                                WHERE key = %(key)s
+                                """
+                            )
+                        ).format(jobs_table=self.jobs_table),
+                        {"key": key},
+                    )
+                    continue
+                if job.status == Status.QUEUED:
+                    await asyncio.sleep(0.1)
```
tobymao, line 326: we can avoid this hack if there's a lock between dequeue and processing can we make that happen easily in postgres?

saq/queue/postgres.py
vchan: It's theoretically possible the job column could be null, but there's no code path to that situation. Maybe if someone logs into the database and updates the job table manually. Want me to remove this? And/or add a not null constraint on the job column?

saq/queue/postgres.py
vchan: I think the Postgres queue should already already handle it. I can probably remove the hack. When a job is dequeued, the job table's status column is updated to active in the same transaction by the queue, but the worker is the one that updates the status of the job object (separate transaction). The Postgres queue also grabs an advisory lock associated with the job though, and if the worker/queue's connection dies, the lock is released. The sweeper already takes care of active jobs without advisory locks.

saq/queue/postgres.py
tobymao: yea sure, let's make job not null

saq/queue/postgres.py
```diff
@@ -613,10 +620,11 @@ async def _dequeue(self) -> list[Job]:
...
             )
             results = await cursor.fetchall()
-        for result in results:
-            job = self.deserialize(result[0])
-            if job:
-                jobs.append(job)
+            for result in results:
+                job = self.deserialize(result[0])
+                if job:
+                    await self.update(job, status=Status.ACTIVE, connection=conn)
```
vchan, line 626: Update the job's status to match the table column status in the same transaction to keep them in sync.

tobymao (approved):

PR #155, opened 2024-09-15
Fix: Handle aborting a job before it has started processing in Postgres

vchan:
If a job hasn't been dequeued yet and it is aborted, its status should be updated to `aborted`.

### DIFF
--- a/saq/queue/postgres.py
+++ b/saq/queue/postgres.py
@@ -504,9 +504,10 @@ class PostgresQueue(Queue):
                 },
             )
             result = await cursor.fetchone()
-            await self.update(job, status=Status.ABORTING, connection=conn)
-        if result == (Status.QUEUED,):
-            await job.finish(Status.ABORTED, error=error)
+            if result == (Status.QUEUED,):
+                await self.finish(job, Status.ABORTED, error=error, connection=conn)
+            else:
+                await self.update(job, status=Status.ABORTING, connection=conn)
 
     async def dequeue(self, timeout: float = 0) -> Job | None:
         """Wait on `self.cond` to dequeue.
@@ -619,10 +620,13 @@ class PostgresQueue(Queue):
         *,
         result: t.Any = None,
         error: str | None = None,
+        connection: AsyncConnection | None = None,
     ) -> None:
         key = job.key
 
-        async with self.pool.connection() as conn, conn.cursor() as cursor:
+        async with self.nullcontext(
+            connection
+        ) if connection else self.pool.connection() as conn, conn.cursor() as cursor:
             if job.ttl >= 0:
                 expire_at = seconds(now()) + job.ttl if job.ttl > 0 else None
                 await self.update(job, status=status, expire_at=expire_at, connection=conn)
--- a/saq/queue/postgres.py
+++ b/saq/queue/postgres.py
@@ -484,7 +484,29 @@ class PostgresQueue(Queue):
 
     async def abort(self, job: Job, error: str, ttl: float = 5) -> None:
         job.error = error
-        await self.update(job, status=Status.ABORTING)
+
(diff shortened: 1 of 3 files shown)


tests/test_queue.py
```diff
@@ -461,14 +461,6 @@ async def test_job_key(self) -> None:
     async def test_schedule(self, mock_time: MagicMock) -> None:
         pass
 
-    async def test_batch(self) -> None:
```
vchan, line 464: This is already in the base test case.

saq/queue/postgres.py
```diff
@@ -484,7 +484,29 @@ async def iter_jobs(
...
+                    jobs_table=self.jobs_table,
+                ),
+                {
+                    "key": job.key,
+                },
+            )
+            result = await cursor.fetchone()
+            await self.update(job, status=Status.ABORTING, connection=conn)
+        if result == (Status.QUEUED,):
+            await job.finish(Status.ABORTED, error=error)
```
tobymao, line 509: so this is all done in a transaction right? even job.finish? why didn’t you add a test? is this code tested? why did you delete one?

saq/queue/postgres.py
```diff
@@ -484,7 +484,29 @@ async def iter_jobs(
...
+
+        async with self.pool.connection() as conn, conn.cursor() as cursor:
+            await cursor.execute(
+                SQL(
+                    dedent(
+                        """
+                        SELECT status
+                        FROM {jobs_table}
+                        WHERE key = %(key)s
+                        FOR UPDATE
```
tobymao, line 496: do we use for update everywhere else? is this needed given the whole
block is in a transaction?

tobymao: why don’t we have for update in sweep?

saq/queue/postgres.py
vchan: There's a test for abort but it didn't test this edge case. I've updated the test now. As for the deleted test, there's already a base test case that's testing batch, so it wasn't necessary. It was also wrong, which is what prompted this fix.

saq/queue/postgres.py
vchan: I used FOR UPDATE here because if a job that hasn't been dequeued yet is aborted, I didn't want dequeue to accidentally pick it up in between the select and the update. As for sweep, do you think we should? I figured if a job's stuck or orphaned, there's no need to lock it. And if it's not, we don't want to lock them in case a worker needs to update it.

saq/queue/postgres.py
```diff
@@ -484,7 +484,29 @@ async def iter_jobs(
...
+                    )
+                ).format(
+                    jobs_table=self.jobs_table,
+                ),
+                {
+                    "key": job.key,
+                },
+            )
+            result = await cursor.fetchone()
+            await self.update(job, status=Status.ABORTING, connection=conn)
```
tobymao, line 507: why isn't this in the same block? should it be

if result == Queued:
  await job.finish(Aborted):
  else:
await self.update(job, ...)

tests/test_queue.py
```diff
@@ -452,6 +452,22 @@ async def asyncTearDown(self) -> None:
         await super().asyncTearDown()
         await teardown_postgres()
 
+    async def get_job_status(self, key: str) -> Status:
```
tobymao, line 455: make this a method in postgres? it's like the same thing right?

saq/queue/postgres.py
```diff
@@ -484,7 +484,13 @@ async def iter_jobs(
 
     async def abort(self, job: Job, error: str, ttl: float = 5) -> None:
         job.error = error
```
tobymao, line 486: you don't need to set this anymore right, you can pass error to update?

vchan: You're right. Updated.

PR #156, opened 2024-09-15
Fix: Retry should not be responsible for releasing a job's lock

vchan:


tobymao:
what was the consequence of this happening? why were there no test failures?

PR #169, opened 2024-10-04
fix: static gzip assets starlette + fallback

### LINKED ISSUE #168: Starlette web not working
Because of: https://github.com/tobymao/saq/commit/8f4aa33c16acea98fa1641da7e914bd224eb94c1, now the starlette website is not loading because can't find pico.min.css or snabbdom.js

image

I'm creating a fix, should we just hardcode that if it's any of the files it gets the .gz path or what approach do you want to take?

Thanks for your work, it's an amazing library :)

aleixrodriala:
Fixes #168 

I have also added the files for fallback in case browser doesn't have gzip compression, maybe not needed.


aleixrodriala:
done

aleixrodriala:
Yes sorry, I'm out now will fix in a couple hours

Issue #172, opened 2024-10-06
Feature: `asyncpg` Queue backend


cofin:
It's nice to see a Postgres backend added.  I'm just starting to experiment with it, and so far, it's great.

I'm curious do you have plans to offer an an `asyncpg` based Queue in addition the current `psycopg` implementation?  If not, would you be willing to accept a PR adding support?

tobymao:
hey @cofin, probably not unless the PR is extremely minimal. i don't think there would be much benefit to it. going to close this issue for now but happy to discuss

cofin:
@tobymao maybe I could ask for your opinion then.  

The LGPL license that `psycopg` has is problematic for my use case (and many other organization have a similar stance). What was the reason for using `psycopg` over `asyncpg`?  Are there additional features (other than sync support) that `psycopg` offers?  

Just to further clarify my use case.  All of the databases I would connect to are Google CloudSQL or AlloyDB instances.  There's a convenient `asyncpg` integration for the auth proxy, but there is no such library for `psycopg`.

tobymao:
@cofin i change my mind. due to it being lgpl, i'll accept a PR that replaces psycopg with it. will you make a PR? i'm  currently quite constrained so may not be able to do it

cofin:
@tobymao, Thanks for reconsidering. I don't mind creating the PR, and I've actually made a bit of progress on it already.  Did you want me to remove the psycopg implementation or allow for either to be used?

tobymao:
probably remove it. is it a pretty easy switch?

cofin:
> probably remove it. is it a pretty easy switch?

Yes, for the most part.  

The main changes are around prepared statements and the way the listen/notify mechanism works.  I'll submit the PR with them both added, you can weigh in once more.  At that point, I can remove the psyocpg adapter before we merge.

PR #177, opened 2024-10-13
feat: `asyncpg` adapter

cofin:
This PR implements the `asyncpg` adapter mentioned in #172 

I've left the current `psycopg` adapter in as a reference, but I can remove (if needed) based on feedback here.


saq/queue/postgres_asyncpg.py
```diff
@@ -0,0 +1,685 @@
...
+class PostgresAsyncpgQueue(Queue):
+    """
+    Queue is used to interact with Postgres using asyncpg.
+    """
+
+    @classmethod
+    def from_url(  # pyright: ignore[reportIncompatibleMethodOverride]
+        cls: type[PostgresAsyncpgQueue],
+        url: str,
+        min_size: int = 4,
```
cofin, line 56: There's no obvious way to resize a queue in `asyncpg`.  Instead of having `min_size` and `max_size` as an init method, the are named arguments here.

saq/queue/postgres_asyncpg.py
```diff
@@ -0,0 +1,685 @@
...
+    def __init__(self, pool: Pool, key: str) -> None:
+        super().__init__()
+        self.pool = pool
+        self.key = key
+        self._connection: PoolConnectionProxy | None = None
+
+    async def _start(self) -> None:
+        if self._connection is None:
+            self._connection = await self.pool.acquire()
+        txn = self._connection.transaction()
```
cofin, line 666: Leaving this as a context manager as it was in the `psycopg` implementation caused the connection to be released too early

cofin: There's no way to disable `autocommit` at the driver level in `asyncpg`; creating the transaction manually.

tobymao:
i just merged my commit, can you rebase?

tobymao:
i can't really compare the two implementations like this, can you redo it so that you just rewrite the existing postgres adapter?

cofin:
> i can't really compare the two implementations like this, can you redo it so that you just rewrite the existing postgres adapter?

Yep, can do.  I'll also bring in your updated changes.

saq/queue/postgres.py
```diff
@@ -61,11 +63,7 @@ class PostgresQueue(Queue):
         stats_table: name of the Postgres table SAQ will write stats to (default "saq_stats")
         dump: lambda that takes a dictionary and outputs bytes (default `json.dumps`)
         load: lambda that takes str or bytes and outputs a python dictionary (default `json.loads`)
-        min_size: minimum pool size. (default 4)
-            The minimum number of Postgres connections.
-        max_size: maximum pool size. (default 20)
-            If greater than 0, this limits the maximum number of connections to Postgres.
-            Otherwise, maintain `min_size` number of connections.
```
cofin, line 68: Moved to the `from_url` method since it can't be modified after initial creation.

cofin:
I've made the updates for the group and priority settings.  I think it's ready for you to take a look at this point.

tests/test_queue.py
```diff
@@ -664,58 +653,49 @@ async def test_finish_ttl_positive(self, mock_time: MagicMock) -> None:
...
-                ).format(self.queue.jobs_table),
-                (job.key,),
+                WHERE key = $1
+                """.format(self.queue.jobs_table),
+                job.key,
             )
-            result = await cursor.fetchone()
             self.assertIsNone(result)
 
     async def test_cron_job_close_to_target(self) -> None:
```
cofin, line 701: This test is a bit flaky.  It passes about 50-60% of the time for me.

saq/queue/postgres.py
```diff
@@ -386,131 +349,97 @@ async def listen(
...
-                          job = %(job)s
-                          ,status = %(status)s
-                          ,scheduled = %(scheduled)s
-                          {expire_at}
-                        WHERE key = %(key)s
-                        """
-                    )
-                ).format(
-                    jobs_table=self.jobs_table,
-                    expire_at=SQL(",expire_at = %(expire_at)s" if expire_at != -1 else ""),
```
cofin, line 419: This is cool syntax sugar I wasn't aware of from `psycopg`.  

Since there isn't a direct equivalent to this, the easiest thing to do was to break this into 2 statements.

tobymao:
benchmarks like 8.5x slower before this is considered, it needs to be fixed benchmarks/simple.py saq_pg

tobymao:
i'm also not a big fan of the api of asyncio compared to psycopg, looks like it's not dbapi. 

given the benchmarks, i'm so far not inclined to accept it, considering lgpl v3 only means users cannot modify the source, it doesn't seem to be that detrimental, but i do understand some companies cannot adhere to it.

cofin:
That's quite a surprising difference on performance.  I definitely would not have expected that great of a difference.

Re: the dbapi comment, you are correct.  It does not have a dbapi compatible interface.  Given this difference, do you want me to spend time investigating the performance gap?

tobymao:
@cofin i've been thinking about it a bit and i apologize but i don't think i have much appetite for this at the moment, so i think we can close this for now. sorry about that and thanks for the effort.

PR #179, opened 2024-10-18
Feat: Support burst mode

izeigerman:


saq/worker.py
```diff
@@ -102,6 +107,19 @@ def __init__(
         self.tasks: set[Task[t.Any]] = set()
         self.job_task_contexts: dict[Job, JobTaskContext] = {}
         self.dequeue_timeout = dequeue_timeout
+        self.burst = burst
+        self.max_burst_jobs = max_burst_jobs
+        self.burst_jobs_processed = 0
+        self.burst_jobs_processed_lock = threading.Lock()
```
tobymao, line 113: this needs to be an async lock, not a thread lock i think

tobymao:
i was thinking we'd change the self.start method to have a different flow with process instead of reusing it.

but maybe this is ok

tobymao (approved):

saq/worker.py
izeigerman: The `process` method is a sync one. Discussed in person

PR #188, opened 2024-11-18
fix job info always full

N1LEX:
__repr__ info of job is always True


tobymao:
you can just use info for this

N1LEX:
@tobymao

PR #196, opened 2024-12-12
Add handling for Redis timeout error in dequeue method

AIvashov:


saq/queue/redis.py
```diff
@@ -335,24 +336,28 @@ async def abort(self, job: Job, error: str, ttl: float = 5) -> None:
...
+                    "RIGHT",
+                )
+        except RedisTimeoutError:
+            logger.debug("Dequeue timed out")
+            return None
+
         if job_id is not None:
             return await self._get_job_by_id(job_id)
 
-        logger.debug("Dequeue timed out")
```
tobymao, line 355: this is no longer handle anymore right

tobymao:
when does this happen? can you give me a little more context

PR #199, opened 2025-01-06
feat!: add metadata for workers

benfdking:
- add a table to store optional metadata for workers such that workers can report things like IP address and other metadata
- add an optional ability for workers to add metadata to that table

**Breaking** 

- in info object, moved stats for a particular worker under stats, which sits side by side now with metadata than straight inside a worker

{ "worker_id": ...stats, ... } -> { "worker_id": { "stats": ...stats, "metadata": ...metadata }, ... }

### DIFF
--- a/saq/types.py
+++ b/saq/types.py
@@ -45,12 +45,23 @@ class JobTaskContext(t.TypedDict, total=False):
     "If this task has been aborted, this is the reason"
 
 
+class WorkerInfo(t.TypedDict):
+    """
+    Worker Info
+    """
+
+    queue_key: t.Optional[str]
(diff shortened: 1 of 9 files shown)


saq/types.py
```diff
@@ -93,6 +102,8 @@ class TimersDict(t.TypedDict):
     "How often to clean up stuck jobs in seconds (default 60)"
     abort: int
     "How often to check if a job is aborted in seconds (default 1)"
+    metadata: int
```
izeigerman, line 105: Why do we need to store it periodically? Isn't it immutable within the worker's lifetime?

PR #200, opened 2025-01-07
fix: http workers pass their own id

benfdking:
**FIRST COMMIT**

Currently, stats are written to the store by the Queue class, and before a Queue instance contains a UUID used to identify the worker.

When a Postgres or Redis Queue is created inside a worker, the ID will be unique to that worker, and the stats are stored as expected. In this case, a Queue instance is created with a unique ID wrapped in a Worker class, which then pushes the stats to the store with the ID from that queue instance, which is unique to that worker. 

This is currently not quite the case with HTTP. In an HTTP Proxy scenario, the proxy creates a Queue instance that contains a unique ID for the worker, as is done in the case of Redis and Postgres. When an HTTP worker Queue is created, this instance will contain the unique ID representative of that worker. When, however, the HTTP worker reports its stats, these, rather than being associated with the ID associated with the worker, are related to the ID of the queue instance in the proxy server.

This allows the stats to be attributed to the worker id rather than the id in the proxy. 

**SECOND COMMIT**

In the second commit, I corrected what I think is a semantic error: a queue should not have a worker_id, currently stored under `uuid` in a Queue instance. The second commit removes this id and puts it inside a Worker instance.

### DIFF
--- a/saq/types.py
+++ b/saq/types.py
@@ -63,9 +63,9 @@ class QueueInfo(t.TypedDict):
     "A truncated list containing the jobs that are scheduled to execute soonest"
 
 
-class QueueStats(t.TypedDict):
+class WorkerStats(t.TypedDict):
     """
-    Queue Stats, could also be used for Worker Stats
+    Worker Stats
     """
 
     complete: int
--- a/saq/types.py
+++ b/saq/types.py
@@ -65,7 +65,7 @@ class QueueInfo(t.TypedDict):
 
 class QueueStats(t.TypedDict):
     """
-    Queue Stats
+    Queue Stats, could also be used for Worker Stats
     """
 
     complete: int
--- a/saq/worker.py
+++ b/saq/worker.py
@@ -36,7 +36,7 @@ if t.TYPE_CHECKING:
         ReceivesContext,
         SettingsDict,
         TimersDict,
-        QueueStats,
+        WorkerStats,
     )
 
 
@@ -118,7 +118,7 @@ class Worker:
         self.burst_jobs_processed = 0
         self.burst_jobs_processed_lock = threading.Lock()
         self.burst_condition_met = False
-        self.id = id if id is not None else uuid1()
+        self.id = uuid1() if id is None else id
 
         if self.burst:
             if self.dequeue_timeout <= 0:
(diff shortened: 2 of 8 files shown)


saq/worker.py
```diff
@@ -115,6 +118,7 @@ def __init__(
         self.burst_jobs_processed = 0
         self.burst_jobs_processed_lock = threading.Lock()
         self.burst_condition_met = False
+        self.id = id if id is not None else uuid1()
```
tobymao, line 121: nit: uuid1() if id is None else id

tests/test_http_stats.py
```diff
@@ -0,0 +1,91 @@
...
+async def echo(_ctx: Context, *, a: int) -> int:
+    return a
+
+
+class ProxyRequestHandler(BaseHTTPRequestHandler):
+    def __init__(self, *args, proxy=None, **kwargs):
+        self.proxy = proxy
+        super().__init__(*args, **kwargs)
+
+    def do_POST(self):
```
tobymao, line 23: nit all lower

izeigerman (approved):

tests/test_http_stats.py
```diff
@@ -0,0 +1,91 @@
...
+
+    async def test_http_proxy_with_two_workers(self) -> None:
+        queue = await create_postgres_queue()
+        proxy = HttpProxy(queue=queue)
+
+        server = HTTPServer(
+            ("localhost", 8080),
+            lambda *args, **kwargs: ProxyRequestHandler(*args, proxy=proxy, **kwargs),
+        )
+        server_thread = threading.Thread(target=server.serve_forever)
```
tobymao, line 52: do we do this anywhere else? i think the general pattern is to use asyncio web app right?

saq/types.py
```diff
@@ -65,7 +65,7 @@ class QueueInfo(t.TypedDict):
 
 class QueueStats(t.TypedDict):
```
tobymao, line 66: rename this?

saq/worker.py
benfdking: 👍

tests/test_http_stats.py
benfdking: not relevant anymore as moved to aiohttp

tests/test_http_stats.py
benfdking: moved over

saq/types.py
benfdking: done

PR #201, opened 2025-01-08
feat!: adding metadata to workers

benfdking:
- add table to store metadata for workers, queue and ip address
- add ip_address and queue_name that worker is working on
- add optional ability for workers to add metadata to that table
- in info moved stats for a particular worker under stats which sits side by side now with metadata than straight inside a worker

### DIFF
--- a/saq/queue/postgres.py
+++ b/saq/queue/postgres.py
@@ -33,6 +33,7 @@ if t.TYPE_CHECKING:
         LoadType,
         QueueInfo,
         WorkerStats,
+        WorkerInfo,
     )
 
 try:
@@ -49,6 +50,7 @@ ENQUEUE = "saq:enqueue"
 DEQUEUE = "saq:dequeue"
 JOBS_TABLE = "saq_jobs"
 STATS_TABLE = "saq_stats"
+METADATA_TABLE = "saq_worker_metadata"
 
 
 class PostgresQueue(Queue):
@@ -90,6 +92,7 @@ class PostgresQueue(Queue):
         name: str = "default",
         jobs_table: str = JOBS_TABLE,
         stats_table: str = STATS_TABLE,
+        metadata_table: str = METADATA_TABLE,
         dump: DumpType | None = None,
         load: LoadType | None = None,
         min_size: int = 4,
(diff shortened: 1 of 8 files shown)


saq/queue/postgres.py
```diff
@@ -107,6 +110,7 @@ def __init__(
         self.versions_table = Identifier(versions_table)
         self.jobs_table = Identifier(jobs_table)
         self.stats_table = Identifier(stats_table)
+        self.metadata_table = Identifier(metadata_table)
```
tobymao, line 113: this should go into the stats table now that we have migrations

benfdking: I've pushed a branch to the PR in which metadata sits in the same table. 

That said, I am not 100% sure about it. 

We get a small win in performance in avoiding a join, but I don't like the `ttl` behavior. It doesn't quite behave the same way as Redis because you have to share the TTL for metadata and stats. 

My best suggestion would be to merge stats and metadata to avoid this weird behavior. WDYT?

tobymao: what do you mean merge?

benfdking: I mean we merge the code top to bottom, get rid of the two separate jobs/calls/methods and just have one stats/info method both for the retrieving of data (which we already do) but also for the pushing of data.

tobymao: sure sounds good

saq/queue/postgres_migrations.py
```diff
@@ -43,4 +43,17 @@ def get_migrations(
...
         ),
+        (
+            2,
+            [
+                SQL(
+                    dedent("""
+ALTER TABLE {stats_table} ADD COLUMN IF NOT EXISTS metadata JSONB;
+ALTER TABLE {stats_table} ADD COLUMN IF NOT EXISTS queue_key TEXT;
+CREATE INDEX IF NOT EXISTS saq_stats_queue_key_idx ON {stats_table} (expire_at);
+CREATE INDEX IF NOT EXISTS saq_stats_queue_key_idx ON {stats_table} (queue_key);
```
tobymao, line 54: shouldn't index names be unique?

tobymao (approved):

saq/queue/postgres_migrations.py
benfdking: 🦅 👁️

PR #202, opened 2025-01-13
feat: introduce simple migration framework

benfdking:
- introduces simple migration framework to make it possible to change tables later down the lline


saq/queue/postgres.py
```diff
@@ -132,10 +138,50 @@ async def init_db(self) -> None:
             if result and not result[0]:
                 return
 
-            for statement in DDL_STATEMENTS:
+            migrations = return_migrations(
+                migrations_table=self.migrations_table,
+                jobs_table=self.jobs_table,
+                stats_table=self.stats_table,
+            )
+            target_number = len(migrations)
```
izeigerman, line 146: `target_version`?

saq/queue/postgres.py
```diff
@@ -132,10 +138,50 @@ async def init_db(self) -> None:
...
+                        SELECT id FROM {migrations_table}
+                        """
+                        )
+                    ).format(migrations_table=self.migrations_table),
                 )
+                result = await cursor.fetchone()
+            except UndefinedTable:
+                result = None
+            if result is not None:
+                current_migration = result[0]
```
izeigerman, line 161: `current_version`

saq/queue/postgres_migrations.py
```diff
@@ -0,0 +1,51 @@
...
+    migrations_table: Identifier,
+    jobs_table: Identifier,
+    stats_table: Identifier,
+) -> t.List[t.List[Composed]]:
+    return [
+        [
+            SQL(
+                dedent("""
+CREATE TABLE IF NOT EXISTS {migrations_table} (
+    id INT
```
izeigerman, line 17: `version`

saq/queue/postgres_migrations.py
```diff
@@ -0,0 +1,51 @@
+import typing as t
+
+from psycopg.sql import Identifier, SQL, Composed
+from textwrap import dedent
+
+
+def return_migrations(
```
izeigerman, line 7: `get_migrations`

saq/queue/postgres.py
tobymao: don't use len(...), let's use the file number

saq/queue/postgres.py
```diff
@@ -128,14 +132,58 @@ async def init_db(self) -> None:
...
+            );
+                        """)
+                ).format(versions_table=self.versions_table)
+            )
+
+            migrations = get_migrations(
+                jobs_table=self.jobs_table,
+                stats_table=self.stats_table,
+            )
+            target_version = len(migrations)
```
tobymao, line 152: let's get the last one, and infer the version from the file name

tobymao: can we make this either a hash value or something less error prone? or perhaps like manual tagging of versions

izeigerman (approved):

saq/queue/postgres.py
izeigerman: I don't think we can do hash cause we need to have total order

tobymao: he's gonna change it to be manual indices

tobymao (approved):

PR #204, opened 2025-01-25
add timezone to cron scheduler

Graeme22:
Hi, thanks for this cool project!

I want to use SAQ for a trading app, where taking into account New York time (and DST) is very important. Currently there is no way to do this as far as I can tell, but it's a small change!
I am unfamiliar with `unittest`, so I may need some guidance making sure the tests work correctly here and adding new ones.

### DIFF
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -6,10 +6,12 @@ import logging
 import time
 import typing as t
 import unittest
+from datetime import datetime, timedelta, timezone
 from unittest import mock
 
 from aiohttp import web
 from aiohttp.test_utils import AioHTTPTestCase
+from time_machine import travel
 
 from saq.job import CronJob, Job, Status, ACTIVE_STATUSES
 from saq.queue import Queue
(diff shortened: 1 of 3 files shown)


tobymao:
hey @Graeme22, please write some tests in worker.py i'm not sure what kind of guidance you need

Graeme22:
> hey @Graeme22, please write some tests in worker.py i'm not sure what kind of guidance you need

Mainly, I'm not sure why it's failing the current tests, I mocked `datetime.now` and it was failing sometimes and not others. I will upload the tests I was trying tomorrow.

Graeme22:
Hi @tobymao, I added `time-machine` as a dev dependency as it was too tricky to mock all of the different date/time related functions across multiple asyncio tasks. Hopefully this should work for users.

tests/test_worker.py
```diff
@@ -368,35 +370,44 @@ async def test_schedule(self, mock_time: MagicMock) -> None:
...
     @mock.patch("saq.worker.logger")
-    @mock.patch("saq.utils.time")
-    async def test_cron(self, mock_time: MagicMock, mock_logger: MagicMock) -> None:
+    async def test_cron(self, mock_logger: MagicMock) -> None:
         with self.assertRaises(ValueError):
             Worker(
                 self.queue,
                 functions=functions,
-                cron_jobs=[CronJob(cron, cron="x")],
+                cron_jobs=[CronJob(sleeper, cron="x")],
```
tobymao, line 378: why is this function changed?

Graeme22: See line 404, one of the checks is if the job is running which requires it to take some time.

tobymao:
can you fix the build? run style and all the tests

Graeme22:
Tests were already passing, just fixed style

tests/test_worker.py
```diff
@@ -368,35 +370,38 @@ async def test_schedule(self, mock_time: MagicMock) -> None:
...
-            mock_logger.info.assert_any_call("Scheduled %s", ["saq:job:default:cron:cron"])
+        with travel(datetime(2025, 1, 1, 0, 0, 1, tzinfo=timezone.utc), tick=True) as traveller:
+            worker = Worker(
+                self.queue,
+                functions=functions,
+                cron_jobs=[CronJob(sleeper, cron="* 12 * * *", kwargs={"sleep": 5})],
+                cron_tz=timezone(offset=timedelta(hours=-3)),  # 3 hours behind (UTC-3)
+            )
+            await worker.queue.connect()
+            self.assertEqual(await self.queue.count("incomplete"), 0)
```
tobymao, line 389: you removed the queue checks

tests/test_worker.py
```diff
@@ -368,35 +370,38 @@ async def test_schedule(self, mock_time: MagicMock) -> None:
...
+            asyncio.create_task(worker.start())
+            await asyncio.sleep(1)
+            self.assertEqual(await self.queue.count("incomplete"), 1)
+
+            traveller.move_to(datetime(2025, 1, 1, 12, 0, 0, tzinfo=timezone.utc))  # noon UTC
+            await asyncio.sleep(2)
```
tobymao, line 395: any way to make these tests faster? can you sleep for less time? perhaps reduce the cron timer?

tests/test_worker.py
Graeme22: Whoops! I'll add those back in.

PR #212, opened 2025-02-07
fix: prevent multiplexer from sending connections after pool shutdown

tobymao:


barakalon:
This is a little awkward.
I originally added implicit, lazy start right on the multiplexer, but now that depends on queue state.
Maybe we should move the lazy start to the queue? I.e. Queue should have a `listen` method that checks if its closed.

PR #215, opened 2025-02-12
fix!: update defaults to not touching status

tobymao:
prior to this commit, job.update() would also override status, this is problematic when using update() for heartbeats since it would override an abort. now, a job status only gets updated if status is passed in directly.

### DIFF
--- a/saq/queue/postgres.py
+++ b/saq/queue/postgres.py
@@ -21,7 +21,7 @@ from saq.job import (
 from saq.multiplexer import Multiplexer
 from saq.queue.base import Queue, logger
 from saq.queue.postgres_migrations import get_migrations
-from saq.utils import now, now_seconds
+from saq.utils import now_seconds
 
 if t.TYPE_CHECKING:
     from collections.abc import Iterable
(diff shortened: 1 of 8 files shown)


izeigerman (approved):

saq/queue/postgres.py
```diff
@@ -459,19 +459,16 @@ async def listen(
...
-
-        for k, v in kwargs.items():
-            setattr(job, k, v)
+    async def _update(self, job: Job, status: Status | None = None, **kwargs: t.Any) -> None:
+        expire_at = kwargs.pop("expire_at", -1)
+        connection = kwargs.pop("connection", None)
 
         async with self.nullcontext(connection) if connection else self.pool.connection() as conn:
+            if not status:
+                status = await self.get_job_status(job.key, for_update=True, connection=conn)
```
barakalon, line 468: I think there's still a race condition here, if the job is aborted after this call but before the update. So this will still happen (just less frequently).

Not sure if you care about that.

Probably need to push the partial update all the way down to the db update.

tests/test_queue.py
```diff
@@ -435,9 +434,8 @@ async def test_sweep(self, mock_time: MagicMock) -> None:
         job3 = await self.enqueue("test", timeout=1)
         for _ in range(4):
             job = await self.dequeue()
-            job.status = Status.ACTIVE
             job.started = 1000
```
barakalon, line 437: Bit awkward that there are two different ways to update the job - setting attributes or kwargs.

saq/queue/postgres.py
tobymao: isn't that what for_update accomplishes?

tobymao: i can also use a transaction if necessary

tobymao: uses a transaction now, there should be no race condition

barakalon: ah, i didn't realize we select for update

barakalon: (i havent been following postgres queue closely)

tobymao: you should the redis queue is falling out of fashion… no transactions lol

barakalon (approved):

saq/queue/postgres.py
barakalon: airbnb doesn't host postgres, so not a chance haha

PR #220, opened 2025-02-26
chore: add heartbeat test for workers

benfdking:


tests/test_worker.py
```diff
@@ -487,6 +487,38 @@ async def handler(_ctx: Context) -> None:
...
+            dequeue_timeout=0.1,
+            burst=True,
+        )
+
+        asyncio.create_task(self.worker.start())
+
+        job_not_refreshed = await self.queue.enqueue("noop", heartbeat=5)
+        job_refreshed = await self.queue.enqueue("noop", heartbeat=5)
+
+        await asyncio.sleep(3)
```
tobymao, line 507: does the sleep need to be this long?

PR #224, opened 2025-04-02
feat: make swept message overrideable

benfdking:


saq/queue/base.py
```diff
@@ -40,6 +40,12 @@
 logger = logging.getLogger("saq")
 
 
+DEFAULT_SWEPT_JOB_ERROR = "swept"
```
tobymao, line 43: this seems passed on in too many places, can it be reduced?

benfdking: Reduced the places, let me know.

tobymao (approved):

saq/queue/http.py
```diff
@@ -117,7 +117,13 @@ def __init__(
         session_callback: t.Optional[t.Callable[[], t.Awaitable[ClientSession]]] = None,
         **kwargs: t.Any,
     ) -> None:
-        super().__init__(name=name, dump=None, load=None)
+        swept_error_message = kwargs.pop("swept_error_message", DEFAULT_SWEPT_JOB_ERROR)
```
izeigerman, line 120: Should this be a top level argument instead?

izeigerman (approved):

PR #226, opened 2025-04-10
Fix: Prevent repeated connect / disconnect in the postgres queue

izeigerman:


benfdking (approved):

saq/queue/postgres.py
```diff
@@ -142,6 +142,7 @@ def __init__(
         self._listener = ListenMultiplexer(self.pool, self._channel)
         self._dequeue_lock = asyncio.Lock()
         self._listen_lock = asyncio.Lock()
+        self._connected = False
```
tobymao, line 145: i thought we had something like this already

izeigerman: It was removed when we introduced support for custom pools

tobymao (approved):

PR #227, opened 2025-04-23
Feat: Improve Postgres dequeue performance with additional indexes

vchan:
These index probably won't be used by the query planner unless the jobs table has hundreds of thousands of rows, but they do significantly speed up the query when there are that many jobs.

### DIFF
--- a/saq/queue/postgres_migrations.py
+++ b/saq/queue/postgres_migrations.py
@@ -56,4 +56,16 @@ CREATE INDEX IF NOT EXISTS saq_stats_queue_key_idx ON {stats_table} (queue_key);
                 ).format(stats_table=stats_table),
             ],
         ),
+        (
+            3,
+            [
+                SQL(
+                    dedent("""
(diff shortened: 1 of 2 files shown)


saq/queue/postgres_migrations.py
```diff
@@ -56,4 +56,16 @@ def get_migrations(
...
             ],
         ),
+        (
+            3,
+            [
+                SQL(
+                    dedent("""
+CREATE INDEX IF NOT EXISTS saq_jobs_status_queue_group_key_idx ON {jobs_table} (status, queue, group_key);
+CREATE INDEX IF NOT EXISTS saq_jobs_status_queue_priority_scheduled_idx ON {jobs_table} (status, queue, priority, scheduled);
+DROP INDEX saq_jobs_dequeue_idx;
```
vchan, line 66: Replaced by `saq_jobs_status_queue_priority_scheduled_idx` above.

tobymao (approved):

PR #232, opened 2025-05-01
Complete example in README.md

foarsitter:
Added a DBHelper class so the examples can be run as is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,24 @@ environment variables:
 ```
 
 ## Example
+
 ```python
 import asyncio
 
 from saq import CronJob, Queue
+
+
+class DBHelper:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -72,10 +82,10 @@ async def test(ctx, *, a):
     return {"x": a}
 
 async def cron(ctx):
-  print("i am a cron job")
+    print("i am a cron job")
 
 async def startup(ctx):
-    ctx["db"] = await create_db()
+    ctx["db"] = DBHelper()
```
tobymao, line 88: let's make it async

foarsitter: Updated the branch, feel free to modify it to your own beliefs. As long it is a working example I'm happy.

PR #235, opened 2025-05-10
chore: don't dequeue during the listen generator

kelsin:


tobymao (approved):

kelsin:
Not a proper fix, need more research. Will open a new one eventually.

Issue #238, opened 2025-05-30
Feature request / question about poll_interval


HaukurPall:
1. Is there a reason why the `apply` method (on `Queue`) does not expose the `poll_interval` argument?

I've been testing out SAQ as a replacement for Celery and noticed that the "base latency" is generally higher with the default arguments of SAQ.

By "base latency" I mean a scenario in which one submits a job and waits for the result. One can f.ex. do this using `apply` and `map`.

Using `poll_interval=0.0` on the `map` method reduces the base latency to something which is lower than the Celery, but setting it as 0.05 still exceeds Celery's base latency.
Just as an API design question, why is the `poll_interval` not exposed on the `apply` method?

I did not noticed anything "bad" happening by setting the `poll_interval` to `0.0`. What are your thoughts on it?

2. In the readme there is some talk about "no polling". It might need some clarification since the `poll_interval` seems to indicate, well, polling. Is the "no polling" only about consuming jobs, but not about retrieving results?

PR #244, opened 2025-06-23
Add poll_interval argument to apply

HaukurPall:
#238 
Also adds a test case for poll_interval.

I'm not sure about the readme changes, but for my use-case (low latency tasks) the `poll_interval` is a crucial part in understanding how the library works and making it respond faster. The mentions of "avoids polling" in the readme were confusing to me as they do not say that it does not apply to the results.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,9 @@ print(job.results)
 # run a job and return the result
 print(await queue.apply("test", a=2))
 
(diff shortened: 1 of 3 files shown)


tobymao:
@HaukurPall thanks for this, please fix the build, run make style

PR #250, opened 2025-07-03
feat: add retry to http calls

eakmanrq:
Adds retry to http calls. An exception is made for dequeue since it is not currently idempotent.

### DIFF
--- a/saq/queue/http.py
+++ b/saq/queue/http.py
@@ -6,10 +6,12 @@ from __future__ import annotations
 
 import json
 import typing as t
+import asyncio
 
 from saq.errors import MissingDependencyError
 from saq.job import Job, Status
 from saq.queue.base import Queue
+from saq.utils import exponential_backoff
 
 if t.TYPE_CHECKING:
     from collections.abc import Iterable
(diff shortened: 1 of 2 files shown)


saq/queue/http.py
```diff
@@ -138,12 +149,43 @@ async def disconnect(self) -> None:
...
+        if isinstance(exception, (ServerTimeoutError, ClientConnectorError)):
+            return True
+        if isinstance(exception, ClientResponseError):
+            return exception.status >= 500
+        if isinstance(exception, ClientError):
+            return True
+        return False
+
+    async def _retry_on_failure(self, operation_name: str, coro_func: t.Callable, *args: t.Any, **kwargs: t.Any) -> t.Any:
+        non_idempotent_operations = {"dequeue"}
```
tobymao, line 162: make this a constant outside

saq/queue/http.py
```diff
@@ -138,12 +149,43 @@ async def disconnect(self) -> None:
             await self.session.close()
             self.session = None
 
+    def _should_retry_on_exception(self, exception: BaseException) -> bool:
+        if isinstance(exception, (ServerTimeoutError, ClientConnectorError)):
+            return True
+        if isinstance(exception, ClientResponseError):
+            return exception.status >= 500
+        if isinstance(exception, ClientError):
```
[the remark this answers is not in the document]
eakmanrq: I don't believe this is correct since `return exception.status >= 500` means that any `ClientResponseError` will be returned in the prior if check.

saq/queue/http.py
```diff
@@ -138,12 +149,49 @@ async def disconnect(self) -> None:
...
+                await asyncio.sleep(delay)
+
     async def _send(self, kind: str, **kwargs: t.Any) -> str:
         assert self.session
 
-        async with self.session.post(self.url, json={"kind": kind, **kwargs}) as resp:
-            resp.raise_for_status()
-            return await resp.text()
+        async def _perform_request() -> str:
+            assert self.session
```
tobymao, line 189: why is this necessary, doesn't it happen on line 180?

saq/queue/http.py
eakmanrq: updated

tobymao (approved):

PR #255, opened 2025-07-15
Refactor/extract runner from main

koldakov:
I wanted to run SAQ from the code, but faced an issue I need to modify sys.argv, because there are other arguments are passed that I can't modify.

This commit allows to run SAQ from the code directly not relying on args:

`saq.runner.run("path.to.settings")`

Also when I was refactoring I've noticed that port by default is None, but`saq.worker.start` accepts port as int only. It works, because `aiohttp.web.run_app` accepts port as `None`, but I mean it's better if we have everything strictly defined.

### DIFF
--- a/saq/__main__.py
+++ b/saq/__main__.py
@@ -1,10 +1,8 @@
 import argparse
-import logging
-import multiprocessing
 import os
 import sys
 
-from saq.worker import check_health, start
+from saq.runner import run
 
 
 def main() -> None:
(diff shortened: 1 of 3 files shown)


koldakov:
Sure, I just need some time.
My bad

koldakov:
@tobymao 1 thing I don't get you have a job for python3.9, but there are multiple places in the code where`type | type` is used. How does it work? 🤔

tobymao:
| works in method definitions because of future annotations, but you can't use them inline in the code is my guess

tobymao:
@koldakov any update on when you will be able to work on this?

PR #265, opened 2025-10-02
chore: Improve worker context typing

ditsuke:
Makes `Worker` generic on the context type so that we can statically prove all of its lifecycle and task functions accept the same context type.


examples/simple.py
```diff
@@ -3,14 +3,17 @@
 import time
 
 from saq import CronJob, Queue
+from saq.types import Context, SettingsDict
 
+CtxType = Context
```
tobymao, line 8: doesn't seem like this alias is necessary, let's remove it?

ditsuke: > doesn't seem like this alias is necessary, let's remove it?

You're right, totally unnecessary. Also don't need CtxType, we just use vanilla base `Context` in this example

tobymao:
the build is still broken

ditsuke:
@tobymao ci should be good for this PR now

PR #266, opened 2025-10-02
fix: Allow tasks to stop before cleanup

ditsuke:
## What and why

Moves `shutdown` callback execution to after `worker.stop()`

`shutdown` callbacks are most often used to cleanup shared resources used by tasks, example being database connections, pools etc. Because these callbacks were executed before calling on `worker.stop()` we were encountering an inconsistent state failures from active tasks trying to use resources cleaned up by our `shutdown` callback.

### DIFF
--- a/saq/worker.py
+++ b/saq/worker.py
@@ -194,11 +194,11 @@ class Worker(t.Generic[CtxType]):
         finally:
             logger.info("Worker shutting down")
 
+            await self.stop()
+
             if self.shutdown:
                 for s in self.shutdown:
                     await s(self.context)
-
-            await self.stop()
 
     async def stop(self) -> None:
         """Stop the worker and cleanup."""
--- a/saq/worker.py
+++ b/saq/worker.py
@@ -5,7 +5,6 @@ Workers
 from __future__ import annotations
 
 import asyncio
-from contextlib import AsyncExitStack
 import contextvars
 import logging
 import os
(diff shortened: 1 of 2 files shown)


tobymao:
looks like you broke the build, you'll have to fix it

ditsuke:
Ready for another look

saq/worker.py
```diff
@@ -192,25 +196,37 @@ async def start(self) -> None:
...
-        all_tasks = list(self.tasks)
-        self.tasks.clear()
-        await cancel_tasks(all_tasks)
-
-        if sys.version_info[0:2] < (3, 9):
-            self.pool.shutdown()
-        else:
-            self.pool.shutdown(cancel_futures=True)
+        async with AsyncExitStack() as stack:
+            await stack.enter_async_context(self._stop_lock)
```
tobymao, line 206: i don't understand all of this code 205-227, can you explain it to me?

ditsuke: - Avoid multiple stop jobs running at once with a lock
- Use a flag to avoid `stop()` cleanup running twice, as `stop()` can trigger another invocation of itself by `start()` by the `Worker.event` shutdown signal.
- Use the new-ish `AsyncExitStack` to cleanly collapse the lock context manager and setting `_stopped` flag on exit into a single `async with` scope.
- Just like before, cancel all tasks both async and threadpool
- But now that we're cancelling tasks (and hence stopping the threadpool) before running shutdown callbacks, we need a new threadpool to run any sync callbacks.

ditsuke: Yes of course. To be honest, I can't fault you, I was also taken aback by the exit stack until I got used to reading and using it myself.

tobymao (approved):

PR #259, opened 2025-07-28
Record worker_id in job metadata once a job is picked up

Seemone:
helps solving https://github.com/tobymao/saq/issues/258

Issue #258, opened 2025-07-28
How to handle orphaned jobs


Seemone:
I noticed that unless I set a timeout, when a worker dies its currently active jobs are stuck forever. This is particularly annoying in case of cron jobs that are never run anymore unless the job entry is manually removed from Redis.

Also my use case mixes short jobs with very long ones (possibly hours) and I can't predict the duration, so establishing a proper timeout isn't possible.

I was looking into writing some self healing code to detected these orphaned jobs but I noticed the job itself has no information about the worker handling it. This is needed to check jobs against live workers and clean up orphaned jobs.

Additional information: I am using multiple queues and thus multiple workers.
The idea was to run the above code in a cron job with a timeout so that would self-cleanup and it would cleanup all the other jobs with timeout=0

Am I missing something that would solve this problem more easily?

thanks

Seemone:
https://github.com/tobymao/saq/issues/258 records worker_id in jobs to help with the solution above

PR #274, opened 2025-10-21
Record worker_id in job metadata once a job is picked up

Seemone:
This is a retry of https://github.com/tobymao/saq/pull/259
Original issue https://github.com/tobymao/saq/issues/258

TL;DR: this change ensures that the worker id of the worker processing a job is store in the job itself, allowing for better control of abandoned jobs.


Seemone:
To clarify, the old PR timed out as I was on leave and then on other projects, and I didn't know how to fix the seemingly unrelated test failure. This time, same code, tests pass.

PR #281, opened 2025-11-27
fix: Do not process job; instead re enqueue if shutdown was requested

### LINKED ISSUE #280: Why not re-qneueu/put back jobs if event set
Currently, a task could simply be awaiting on a dequeue for a long time even though SIGTERM has already been sent. And can possibly pick up a job right when it's about to be killed.

We can get around this by setting a dequeue timeout to some minimal number like 5s.

But why not reenqueue the job if you can check that event is set?

vikigenius:
Fixes #280

### DIFF
--- a/saq/worker.py
+++ b/saq/worker.py
@@ -351,6 +351,21 @@ class Worker(t.Generic[CtxType]):
             if job is None:
                 return False
 
+            if self.event.is_set():
+                try:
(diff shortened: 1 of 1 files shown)


vikigenius:
You are right, I don't think it's good to enqueue again and just introduces potential complexity. If we are letting it fail, we don't need to do anything, it will auto fail after timeout anyway I will close the PR

PR #282, opened 2025-12-01
feat: Allow defining non-error exceptions for control flow

ditsuke:
## Summary

In saq, retry policies and exceptions can be used as a form of control flow. For example, internally, we use a `RetryException` to indicate that a task should be retried at a later point. However, right now saq will still log the exception as an error which makes for some log noise. Here I propose to introduce a way to pass a set of exception types
that saq will not log as errors.


tobymao:
Since this is only being used for error messaging, I don't see too much value in this. I'm not sure I'll want to bring this patch in.

ditsuke:
> Since this is only being used for error messaging, I don't see too much value in this. I'm not sure I'll want to bring this patch in.

Fair. Would you be more open to a more explicit control flow mechanism instead? Say something like a `SaqDoRetry` return/exception?

tobymao:
closing for now due to inactivity

PR #296, opened 2026-07-16
fix: return scheduled job ids from PostgresQueue.schedule

### LINKED ISSUE #295: Hardcoded empty list in PostgresQueue.schedule
Hi there, thanks for bringen SAQ! 

The `schedule` method behaves differently between RedisQueue and PostgresQueue. The postgres one returns an empty list.

https://github.com/tobymao/saq/blob/dec040aa435b1aa634d18dc42cc458d9a260e868/saq/queue/postgres.py#L353-L356

The desired behaviour would be that the scheduled tasks should be returned. This outputs the correct log messages in worker. 

https://github.com/tobymao/saq/blob/dec040aa435b1aa634d18dc42cc458d9a260e868/saq/worker.py#L272-L276

mlw157:
## What and why

`PostgresQueue.schedule()` unconditionally returned `[]`:

```python
async def schedule(self, lock: int = 1) -> t.List[str]:
    await self._dequeue()
    return []
```

So the worker's scheduled-jobs log never fired on the Postgres backend, unlike Redis:

```python
job_ids = await self.queue.schedule(lock)
if job_ids:
    logger.info("Scheduled %s", job_ids)
```

This left a gap in observability between the Redis and Postgres backends.
## Fix

`schedule()`already delegates its work to` _dequeue()`, which was discarding the
jobs it activated. `_dequeue()` now collects and returns the dequeued `job.ids`, and
`schedule()` returns them, matching `RedisQueue.schedule`, which returns the ready
job ids. Other `_dequeue()` callers ignore the return value, so they're unaffected.

## Tests

Implements the previously-skipped `tests/test_queue.py::TestPostgresQueue.test_schedule`,
asserting `schedule()` returns the ids of the jobs it moves to active (mirroring the
Redis test_schedule).

Closes #295

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -54,4 +54,6 @@ jobs:
         pip install -e .[dev,web]
     - name: Run Checks
       run: |
-        make style test
+        make style
+        # abort with a full stack dump of all threads if the suite hangs
+        PYTHONFAULTHANDLER=1 timeout --signal=ABRT 600 python -m coverage run -m unittest -vv
--- a/saq/queue/postgres.py
+++ b/saq/queue/postgres.py
@@ -152,7 +152,8 @@ class PostgresQueue(Queue):
     async def init_db(self) -> None:
         async with self.pool.connection() as conn, conn.cursor() as cursor, conn.transaction():
             await cursor.execute(
-                SQL("SELECT pg_try_advisory_lock(%(key1)s, 0)"),
+                # transaction-scoped so the lock can't outlive init_db on a pooled connection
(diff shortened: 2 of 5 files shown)


tobymao:
Thanks for the fix! I pushed a small commit on top: RedisQueue.schedule only reports jobs whose scheduled time arrived (the Lua script's ZRANGEBYSCORE starts at 1, excluding unscheduled jobs), so I filtered the returned ids to jobs with an explicit scheduled time — otherwise plain enqueued jobs claimed by the schedule tick would get logged as "Scheduled [...]". Also extended test_schedule with an unscheduled job to cover the distinction, mirroring the Redis test. Merging once CI is green.

